PR #1, opened 2020-07-21
allow leader election to accept namespace

jmrodri:
* remove isRunLocal()
* change errNoNamespace value, it now returns
  "Namespace is required when running outside the cluster"
* update tests

### DIFF
--- a/leader/leader.go
+++ b/leader/leader.go
@@ -31,16 +31,6 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-type runModeType string
-
-const (
-	localRunMode runModeType = "local"
-)
-
-// forceRunModeEnv indicates if the operator should be forced to run in either local
-// or cluster mode (currently only used for local mode)
-var forceRunModeEnv = "OSDK_FORCE_RUN_MODE"
-
 // errNoNamespace indicates that a namespace could not be found for the current
 // environment
 var errNoNamespace = fmt.Errorf("Namespace is required when running outside the cluster")
--- a/leader/leader.go
+++ b/leader/leader.go
@@ -43,11 +43,10 @@ var forceRunModeEnv = "OSDK_FORCE_RUN_MODE"
 
 // errNoNamespace indicates that a namespace could not be found for the current
 // environment
-var errNoNamespace = fmt.Errorf("namespace not found for current environment")
+var errNoNamespace = fmt.Errorf("Namespace is required when running outside the cluster")
 
-// errRunLocal indicates that the operator is set to run in local mode (this error
-// is returned by functions that only work on operators running in cluster mode)
-var errRunLocal = fmt.Errorf("operator run mode forced to local")
+// namespaceDir is the default location where namespace information is stored
+var namespaceDir = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
 // podNameEnvVar is the constant for env variable POD_NAME
 // which is the name of the current pod.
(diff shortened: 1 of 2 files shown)


leader/leader.go
```diff
@@ -44,11 +44,10 @@ var forceRunModeEnv = "OSDK_FORCE_RUN_MODE"
```
camilamacedo86, line 44: IMO all this stuff will be removed since for the new layout we are using the controller-runtime implementation instead of that.

leader/leader.go
```diff
@@ -60,23 +59,22 @@ var log = logf.Log.WithName("leader")
...
+// namespace. If run outside a cluster, it will return an error. It
 // continuously tries to create a ConfigMap with the provided name and the
 // current pod set as the owner reference. Only one can exist at a time with
 // the same name, so the pod that successfully creates the ConfigMap is the
-// leader. Upon termination of that pod, the garbage collector will delete the
-// ConfigMap, enabling a different pod to become the leader.
-func Become(ctx context.Context, lockName string) error {
+// leader. Upon termination of that pod, the garbage collector will delete
+// the ConfigMap, enabling a different pod to become the leader.
+func Become(ctx context.Context, lockName string, ns string) error {
```
joelanford, line 69: WDYT of duplicating the [leaderelection.Options](https://github.com/kubernetes-sigs/controller-runtime/blob/229c3c357d9e6b07c3d6774010c35161a82b08f9/pkg/leaderelection/leader_election.go#L34-L47) struct from controller-runtime and using that as a parameter to `leader.Become()`

```suggestion
func Become(ctx context.Context, opts Options) error {
```

leader/leader.go
jmrodri: @camilamacedo86 it won't be removed because this is a library for operator developers. This offers leader for life. While it won't be used by the ansible or helm directly, it can be used by other developers.

leader/leader.go
```diff
@@ -210,10 +198,7 @@ func isPodEvicted(pod corev1.Pod) bool {
 
 // getOperatorNamespace returns the namespace the operator should be running in.
 func getOperatorNamespace() (string, error) {
```
estroz, line 200: Nit: have this take a path to make it testable (without resetting a global in tests)
```suggestion
func getOperatorNamespace(namespacePath string) (string, error) {
```

mhrivnak:
This approach to leader election depends on a "lease" resource, implemented with a ConfigMap, having an owner reference to the Pod that the operator is running in. If the Pod goes away, the lease automatically goes with it, and thus someone else can become the leader.

But given that, I don't see how this is going to work if the operator itself is either running outside the cluster, or running in a different namespace than where the lease is. Mind elaborating on the goal and how it's going to work?

joelanford:
@mhrivnak Yeah, that's totally right. I had lost track of that constraint. Given that, it probably doesn't make much sense to accept an explicit namespace since this will only ever work in-cluster, and in-cluster we can always get the pod namespace from the mounted service account secret.

Do you think it's reasonable to use the pod hostname if `POD_NAME` isn't set (as long as we document that the pod spec `hostname` value should not be overridden)?

Or should we keep the hard requirement on `POD_NAME`?

mhrivnak:
> Do you think it's reasonable to use the pod hostname if `POD_NAME` isn't set (as long as we document that the pod spec `hostname` value should not be overridden)?
> 
> Or should we keep the hard requirement on `POD_NAME`?

My first implementation used the pod hostname, but that was [changed](https://github.com/operator-framework/operator-sdk/pull/617) after it was [pointed out as not being a reliable way](https://github.com/operator-framework/operator-sdk/pull/617#issuecomment-430268001) to identify the pod name. Based on the comments, it sounds like the hostname should be avoided.

leader/leader.go
jmrodri: @camilamacedo86 I apologize you are correct the `forceRunModeEnv` and other unused variables can go. They were removed.

jmrodri:
Based on the comments above, it sounds like it would be a bad idea to accept a namespace parameter. Closing this PR.

PR #7, opened 2020-07-22
Leader remove runlocal

jmrodri:


leader/leader.go
```diff
@@ -71,7 +60,7 @@ func Become(ctx context.Context, lockName string) error {
 
 	ns, err := getOperatorNamespace()
 	if err != nil {
-		if err == errNoNamespace || err == errRunLocal {
+		if err == errNoNamespace {
 			log.Info("Skipping leader election; not running in a cluster.")
 			return nil
 		}
```
joelanford, line 66: I think if I was using this leader election library, I would want the ability to handle the no namespace error myself. WDYT of exporting `ErrNoNamespace` so that callers can do:
```go
if err := leader.Become(...); err != nil {
	if errors.Is(err, leader.ErrNoNamespace) {
		// handle the no namespace error
	}
	// other error handling
}
```

And then just having this check return the error directly.

leader/leader.go
```diff
@@ -210,10 +199,7 @@ func isPodEvicted(pod corev1.Pod) bool {
 
 // getOperatorNamespace returns the namespace the operator should be running in.
 func getOperatorNamespace() (string, error) {
-	if isRunModeLocal() {
-		return "", errRunLocal
-	}
-	nsBytes, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	nsBytes, err := ioutil.ReadFile(defaultNamespaceLocation)
```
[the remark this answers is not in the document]
jmrodri: That's an interesting approach. Thanks.

leader/leader.go
```diff
@@ -31,72 +31,89 @@ import (
...
 // attempts to become the leader.
 const maxBackoffInterval = time.Second * 16
 
+type Option func(*Config) error
+
+type Config struct {
+	Client crclient.Client
+}
+
+func (c *Config) SetDefaults() error {
```
joelanford, line 58: Users don't need this since there's nothing they can really do with a `leader.Config` struct other than use `WithClient`, and even that doesn't require them to deal with a `Config` struct in their code.

I think it should be unexported.

In fact, I think we could unexport `Config` as well. And leave just `Option` and `WithClient` exported.

leader/leader_test.go
```diff
@@ -0,0 +1,213 @@
...
+		It("should trim whitespace from namespace", func() {
+
+			nsFile, err := setupNamespace("   testnamespace	   ")
+			if err != nil {
+				Fail(err.Error())
+			}
+			defer os.Remove(nsFile.Name())
+			readNamespace = func() ([]byte, error) {
+				return ioutil.ReadFile(nsFile.Name())
+			}
```
joelanford, line 128: This can be simplified to:

```go
readNamespace = func() ([]byte, err) {
    return []byte("   testnamespace    "), nil
}
```

jmrodri: Oh man not sure how I missed that. Thanks

leader/leader_test.go
```diff
@@ -0,0 +1,187 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("Leader", func() {
+
```
joelanford, line 16: I think this `Describe` layer is unnecessary. Every test in this suite (i.e. package) is a leader test.

leader/leader.go
```diff
@@ -27,76 +27,93 @@ import (
...
+	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+}
+
 var log = logf.Log.WithName("leader")
 
 // maxBackoffInterval defines the maximum amount of time to wait between
 // attempts to become the leader.
 const maxBackoffInterval = time.Second * 16
 
+type Option func(*config) error
```
joelanford, line 52: Ah crap. I totally messed this up. Now that I'm seeing this, I feel like it's a little user-hostile. If we don't export `Config`, users can't provide their own `Option` implementations.

Given that we only have one field to set, it might not seem important now, but if we add more fields in the future, we want to give users the flexibility to write their own `Option`s. 

Opening this up later would be a breaking change to the `Option` definition.

So given all of this, I think we want:
```go
type Option func(*Config) error

type Config struct { ... }

// now that Config is exported, setDefaults would need 
// to be unexported since this only needs to be called
// by us after all options have run.
func (c *Config) setDefaults() error { ... }
```

Sorry for the back and forth!

joelanford:
Following up about controller-runtime manager integration.

It doesn't seem immediately do-able given controller-runtime's direct use of [`resourcelock.Interface`](https://godoc.org/k8s.io/client-go/tools/leaderelection/resourcelock#Interface), which seems to have a hardcoded expectation of using leases.

Here's where the manager starts the pre-leader-election runnables vs the post-leader-election runnables: https://github.com/kubernetes-sigs/controller-runtime/blob/ab02953d0bcb6467333fcc5809eb1df83a8d89c4/pkg/manager/internal.go#L465-L478.

The only way I can see this working would be for controller-runtime to give callers a hook into [`startLeaderElection`](https://github.com/kubernetes-sigs/controller-runtime/blob/ab02953d0bcb6467333fcc5809eb1df83a8d89c4/pkg/manager/internal.go#L602-L639) to override the default. Perhaps it could abstract the leader election implementation to something that allows a [`LeaderCallbacks`](https://github.com/kubernetes/kubernetes/blob/8a3fc2d7e2043bc980bb5301d2f4292020f1a2dd/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L164-L173) to be injected and implements `Run(context.Context)`. So perhaps:

```go
type LeaderElector interface {
    Run(context.Context)
}
```

And then a new manager Option field:

```go
type Option struct {
	// NewLeaderElector overrides the default lease-based leader elector
	// and allows callers to use their own leader election implementation.
	NewLeaderElector func(leaderelection.LeaderCallbacks) LeaderElector
}
```

This would definitely require a controller-runtime proposal and thorough discussion, so not anything that could happen anytime really soon.

leader/leader.go
```diff
@@ -27,76 +27,93 @@ import (
...
+
 // Become ensures that the current pod is the leader within its namespace. If
 // run outside a cluster, it will skip leader election and return nil. It
 // continuously tries to create a ConfigMap with the provided name and the
 // current pod set as the owner reference. Only one can exist at a time with
 // the same name, so the pod that successfully creates the ConfigMap is the
 // leader. Upon termination of that pod, the garbage collector will delete the
 // ConfigMap, enabling a different pod to become the leader.
-func Become(ctx context.Context, lockName string) error {
+func Become(ctx context.Context, lockName string, opts ...Option) error {
```
camilamacedo86, line 88: We repaced the sdk impl for the controller-runtime one. So, I do not think that we should keep it at all. 
@joelanford why we will keep our implementation since all opers now are using the controller-runtime one? 
I indeed update the docs : https://github.com/operator-framework/operator-sdk/pull/3514 so I am looking forward to doing know if we should or not remove it from docs as well.

jmrodri: @camilamacedo86 my understanding was that this is a leader for life and we kept it for someone that wanted to have leader for life. It will not be used in the SDK but could be used by an operator developer.

leader/leader_test.go
```diff
@@ -0,0 +1,187 @@
...
+		})
+		It("should not return an error", func() {
+			os.Setenv("POD_NAME", "leader-test")
+			readNamespace = func() ([]byte, error) {
+				return []byte("testns"), nil
+			}
+
+			err := Become(context.TODO(), "leader-test", WithClient(client))
+			Expect(err).Should(BeNil())
+		})
```
joelanford, line 64: Can we add a test case for returning `ErrNoNamespace` from `Become` and using `errors.Is(err, ErrNoNamespace)`?

leader/leader_test.go
jmrodri: If I remove it, it doesn't seem to work. I renamed it to `Leader election` to better describe it. I would say, it works so I'd rather not change this.

leader/leader.go
joelanford: Still want this function to be unexported though, I think:
```suggestion
func (c *Config) setDefaults() error {
```

leader/leader.go
jmrodri: Reverted the commit that unexported Config. I then added a new commit to unexport just `setDefaults`

leader/leader.go
mhrivnak: A lot of existing operators are using the leader-for-life implementation. They'll either need a way to continue using that, or will need an easy path to migrating to the lease-based election.

These days the lease-based election is easy to use and robust, which hasn't always been the case. But even still, each pattern has advantages and disadvantages.

PR #9, opened 2020-07-24
move handler*test.go -> enqueue_annotation*test.go

jmrodri:


jmrodri:
This might not be the right solution.

PR #14, opened 2020-07-26
Add check-license target

jmrodri:


Makefile
```diff
@@ -23,7 +23,23 @@ test: ## Run unit tests
...
 	@go vet $(addprefix ./, $(SOURCE_DIRS))
 
-check: fmtcheck vet lint build test ## Pre-flight checks before creating PR
+tidy: ## Tidy go dependencies
+	@go mod tidy
+
+check-license: $(SOURCES)
+	@./hack/check-license.sh "$(SOURCES)"
+
+fake:
```
camilamacedo86, line 32: why we need the fake one?

jmrodri: because I was experimenting and forgot to delete it :)

PR #15, opened 2020-07-26
Add test for InstrumentedEnqueueRequestForObject

jmrodri:
Added a test for the `InstrumentedEnqueueRequestForObject` to increase the code coverage.

Before the test, handler was at 71.2%
```
$ make test                                                          
ok      github.com/operator-framework/operator-lib/handler      1.080s  coverage: 71.2% of statements   
...
```

After adding the test, coverage goes to 100%
```
$ make test                                                  
ok      github.com/operator-framework/operator-lib/handler      1.082s  coverage: 100.0% of statements
...
```

### DIFF
--- a/handler/handler_suite_test.go
+++ b/handler/handler_suite_test.go
@@ -19,10 +19,9 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 )
 
 func TestEventhandler(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecsWithDefaultAndCustomReporters(t, "Handler Suite", []Reporter{printer.NewlineReporter{}})
+	RunSpecs(t, "Handler Suite")
 }
--- a/handler/instrumented_enqueue_object_test.go
+++ b/handler/instrumented_enqueue_object_test.go
@@ -1,0 +1,149 @@
+// Copyright 2020 The Operator-SDK Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handler
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllertest"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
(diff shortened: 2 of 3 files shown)


handler/handler_suite_test.go
```diff
@@ -19,10 +19,9 @@ import (
```
camilamacedo86, line 19: just a nit .. Add the sufix _test to the package. ( handler_test )

camilamacedo86: reason:  the suffix `_test` which has [special meaning for the Go compiler](https://github.com/golang/go/issues/25223).

handler/instrumented_enqueue_object_test.go
```diff
@@ -0,0 +1,149 @@
...
+					Name:      pod.Name,
+				},
+			}))
+
+			gauges, err := metrics.Registry.Gather()
+			Expect(err).Should(BeNil())
+			Expect(len(gauges)).To(Equal(1))
+			for _, l := range gauges[0].Metric[0].Label {
+				if l.Name == ptrString("name") || l.Name == ptrString("namespace") {
+					Expect(l.Value).To(Equal("biz"))
```
camilamacedo86, line 127: Why we are using `ptrString("name")`?

handler/instrumented_enqueue_object_test.go
```diff
@@ -0,0 +1,149 @@
...
+				}
+			}
```
camilamacedo86, line 70:
```suggestion
                        name := "name"
                        namespace := "namespace"
			for _, l := range gauges[0].Metric[0].Label {
				if l.Name ==&name || l.Name ==&namespace {
					Expect(l.Value).To(Equal("biz"))
				}
			}
```

camilamacedo86: I do not think that we need the `ptrString` in this case.

camilamacedo86: Also, name should be == namespace? Should  not it be l.Namespace in the second conditional?

handler/instrumented_enqueue_object_test.go
```diff
@@ -0,0 +1,149 @@
...
```
joelanford, line 129: Maybe write a helper function for this and also check the metric's GVK labels and the metric value?

jmrodri:
> Just a few nits:
> 
>     * Could we add the suffix `_test` which has [special meaning for the Go compiler](https://github.com/golang/go/issues/25223)? So, it would be `package handler_test` instead of `package handler`
>       `

No, I wanted it in the same package as the code so I could also test the unexported methods. If we put it in `package handler_test` then we can only test the exported method which is great but sometimes it's easier to test the error conditions of the unexported methods directly.

handler/instrumented_enqueue_object_test.go
```diff
@@ -0,0 +1,149 @@
...
+	var instance InstrumentedEnqueueRequestForObject
+	var pod *corev1.Pod
+
+	BeforeEach(func() {
+		q = controllertest.Queue{Interface: workqueue.New()}
+		instance = InstrumentedEnqueueRequestForObject{}
+		pod = &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "biz",
+				Name:      "biz",
```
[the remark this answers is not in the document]
jmrodri: Excellent idea.

handler/instrumented_enqueue_object_test.go
```diff
@@ -0,0 +1,149 @@
...
+			Expect(len(gauges)).To(Equal(1))
+			for _, l := range gauges[0].Metric[0].Label {
+				if l.Name == ptrString("name") || l.Name == ptrString("namespace") {
+					Expect(l.Value).To(Equal("biz"))
+				}
+			}
+		})
+	})
+
+	Describe("Delete", func() {
```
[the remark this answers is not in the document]
jmrodri: @joelanford I added contexts, excellent tip.

PR #16, opened 2020-07-27
Format source code in go doc.

jmrodri:


handler/enqueue_annotation.go
```diff
@@ -47,33 +47,32 @@ const (
 // of the parent resource which is identified by annotations. `NamespacedNameAnnotation` and
 // `TypeAnnotation` together uniquely identify an owner resource to reconcile.
 //
-// handler.EnqueueRequestForAnnotation can be used to trigger reconciliation of resources which are
-// cross-referenced.  This allows a namespace-scoped dependent to trigger reconciliation of an owner which
-// is in a different namespace, and a cluster-scoped dependent can trigger the reconciliation of a
-// namespace(scoped)-owner.
+//		handler.EnqueueRequestForAnnotation
```
estroz, line 50: Is this supposed to be indented? This name reads like a noun in a sentence.

jmrodri: Honestly I couldn't tell if it was meant to be an example of how to use it and then a description. So I intended it so it would show up as code in the godoc.

image

It being a noun makes sense after re-reading it.

jmrodri: Addressed the above concern.

PR #18, opened 2020-07-30
add lint check for we ensure that all exported code will be documented

camilamacedo86:
**Description**
- Add lint check for we ensure that all exported code will be documented
E..g 

```
leader/leader.go:74:1: exported function `WithClient` should have comment or be unexported (golint)
func WithClient(cl crclient.Client) Option {
```
- Add comments in the exported code that was missing them. 
 
**Motivation**
Ensure that its godocs will be fine :-)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,12 @@
+issues:
+  # don't skip warning about doc comments
+  # don't exclude the default set of lint
+  exclude-use-default: false
+  # restore some of the defaults
+  # (fill in the rest as needed)
+  exclude-rules:
+    - linters: [errcheck]
+      text: "Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*printf?|os\\.(Un)?Setenv). is not checked"
 linters:
   disable-all: true
   enable:
@@ -19,6 +28,5 @@ linters:
     - unused
     - govet
     - typecheck
-
 run:
   deadline: 5m
--- a/leader/leader.go
+++ b/leader/leader.go
@@ -49,8 +49,10 @@ var log = logf.Log.WithName("leader")
 // attempts to become the leader.
 const maxBackoffInterval = time.Second * 16
 
+// Option are the arguments for creating a new Become leader
 type Option func(*Config) error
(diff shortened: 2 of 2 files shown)


.golangci.yml
```diff
@@ -1,3 +1,6 @@
+issues:
+  # don't skip warning about doc comments
+  exclude-use-default: false
```
joelanford, line 3: The `exclude-use-default` setting [ignores a lot more than just go doc comments](https://golangci-lint.run/usage/configuration/). So disabling it means we'll start seeing errors for the godoc comments (good!)... and other lint errors (bad!).

controller-runtime uses the follow setting, which seems like a reasonable middle ground.
 
```suggestion
issues:
  # don't exclude the default set of lint 
  exclude-use-default: false

  # restore some of the defaults
  # (fill in the rest as needed)
  exclude-rules:
  - linters: [errcheck]
    text: "Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*printf?|os\\.(Un)?Setenv). is not checked"

```

leader/leader.go
```diff
@@ -49,8 +49,10 @@ var log = logf.Log.WithName("leader")
 // attempts to become the leader.
 const maxBackoffInterval = time.Second * 16
 
+// Option are the arguments for creating a new Become leader
```
joelanford, line 52:
```suggestion
// Option is a function that can modify Become's Config
```

leader/leader.go
```diff
@@ -49,8 +49,10 @@ var log = logf.Log.WithName("leader")
...
 type Option func(*Config) error
 
+// Config configures the Become leader
```
joelanford, line 55:
```suggestion
// Config defines the configuration for Become
```

leader/leader.go
```diff
@@ -71,6 +73,7 @@ func (c *Config) setDefaults() error {
 	return nil
 }
 
+// WithClient allows use the Become leader with the WithClient an Client informed
```
joelanford, line 76:
```suggestion
// WithClient returns an Option that sets the Client used by Become
```

Issue #24, opened 2020-08-05
Check the node status when leader re-election.


kasonglee:
## Feature Request

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Example: "I have an issue when (...)"

From leader/leader.go, leader re-election works after the default timeout 5-min since the condition
Pod.status.phase == "Failed" && Pod.Status.Reason == "Evicted" when a worker node is failed.
I have an opinion that leader re-election can work almost immediately when the condition contains checking the status of the node where the leader pod is running.

**Describe the solution you'd like**
A clear and concise description of what you want to happen. Add any considered drawbacks.

Check the condition of the node where the leader pod is running with [Node.Type == "NodeReady" && Node.Status != "ConditionTrue"] and when the node has been failed, delete the leaderPod (only mark the pod with 'terminating' because the node where the pod is running has been failed) and the configmap lock whose OwnerReference is leaderPod)

Pictures below are the test of node-check for leader re-election.(test-operator-xxx-xxx**qb** was the leaderPod)

![3](https://user-images.githubusercontent.com/66303838/89374604-d88bef80-d726-11ea-8b56-23f254ef5d58.png)
![4](https://user-images.githubusercontent.com/66303838/89374641-f6f1eb00-d726-11ea-99d5-a58ba91e830c.png)
![5](https://user-images.githubusercontent.com/66303838/89374649-f9ecdb80-d726-11ea-8b45-7dc0e3a3b708.png)

Making --pod-eviction-timeout to be short can be another approach. However, I sure that above approach can bring more reliability since we don't know appropriate time out.

And Is there any drawbacks when making --pod-eviction-timeout to be very very short? 

@HyungJune

mhrivnak:
> And Is there any drawbacks when making --pod-eviction-timeout to be very very short?

The use case for the timeout is that a Node gets temporarily disconnected from the cluster but is then able to re-connect. The shorter the timeout, the more risk that the Node will re-appear and all the workloads on it will try to keep running despite having had their Pods deleted. Probably they would not run for long, but it may take some time for the local kubelet to catch up and stop all the containers.

This scenario greatly benefits from fencing. Ideally you allow a separate component to monitor the Node health, and then take action to ensure that a missing Node will not return before deleting its workloads. This can be done for example by powering off the Node's underlying machine.

kasonglee:
[...]

PR #25, opened 2020-08-06
leader: check the node status for the leader-election (#24)

HyungJune:
node-check for leader re-election #24 

#24 
Check the condition of the node where the leader pod is running with [Node.Type == "NodeReady" && Node.Status != "ConditionTrue"] and when the node has been failed, delete the leaderPod (only mark the pod with 'terminating' because the node where the pod is running has been failed) and the configmap lock whose OwnerReference is leaderPod). 

This approach can bring more reliability

### DIFF
--- a/leader/leader.go
+++ b/leader/leader.go
@@ -167,6 +167,7 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 				log.Info("Leader lock configmap owner reference must be a pod.", "OwnerReference", existingOwners[0])
 			default:
 				leaderPod := &corev1.Pod{}
+				// leaderNode := &corev1.Node{}
 				key = crclient.ObjectKey{Namespace: ns, Name: existingOwners[0].Name}
 				err = config.Client.Get(ctx, key, leaderPod)
 				switch {
@@ -181,6 +182,14 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 					err := config.Client.Delete(ctx, leaderPod)
 					if err != nil {
 						log.Error(err, "Leader pod could not be deleted.")
+					}
+				case isNotReadyNode(ctx, config.Client, leaderPod.Spec.NodeName):
+					log.Info("the status of the node where operator pod with leader lock was running has been 'notReady'")
+					log.Info("Deleting the leader.")
+
+					//Mark the termainating status to the leaderPod and Delete the configmap lock
+					if err := deleteLeader(ctx, config.Client, leaderPod, existing); err != nil {
+						return err
 					}
 
 				default:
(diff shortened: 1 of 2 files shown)


leader/leader.go
```diff
@@ -167,6 +167,7 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 				log.Info("Leader lock configmap owner reference must be a pod.", "OwnerReference", existingOwners[0])
 			default:
 				leaderPod := &corev1.Pod{}
+				// leaderNode := &corev1.Node{}
```
camilamacedo86, line 170: we should not merge  code commented

leader/leader.go
```diff
@@ -167,6 +167,7 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 				log.Info("Leader lock configmap owner reference must be a pod.", "OwnerReference", existingOwners[0])
```
camilamacedo86, line 167: We need to add the logic of the fragments to here as well for we are able to generate the changelog. 

c/c @jmrodri

leader/leader.go
```diff
@@ -182,6 +183,14 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 					if err != nil {
 						log.Error(err, "Leader pod could not be deleted.")
 					}
+				case isNotReadyNode(ctx, config.Client, leaderPod.Spec.NodeName):
+					log.Info("the status of the node where operator pod with leader lock was running has been 'notReady'")
+					log.Info("Deleting the leader.")
+
+					//Mark the termainating status to the leaderPod and Delete the configmap lock
+					if err := deleteLeader(ctx, config.Client, leaderPod, existing); err != nil {
```
camilamacedo86, line 191: Should we not use the same func across all implementation? Also, why is required to remove the configMap?

leader/leader_test.go
```diff
@@ -209,4 +209,137 @@ var _ = Describe("Leader election", func() {
 			Expect(pod.TypeMeta.Kind).To(Equal("Pod"))
 		})
 	})
+
```
camilamacedo86, line 212: PS.: see that is not all changes covered by the tests. its % decreased.

leader/leader.go
jmrodri: There's really no changelog fragment for the operator-lib. The releases are just a signed tag.

camilamacedo86: IMO we should have since we should have a way to know what changed in each version/release.

leader/leader.go
kasonglee: 1) The reason i added the function(delete Leader) is for the lint-test. golangci-lint limits Cyclomatic complexity in the function. so we have to divide the code into multiple functions.
2) if i do not remove the configmap, leader election will not occur imediately when the node where the leader pod is running was 'not ready' status. Because the leader pod was just marked as 'terminating'.(the garbage collector do not delete the configmap until the leader pod is actually deleted)

leader/leader.go
kasonglee: thanks, I will delete the comment at the new commit.

leader/leader_test.go
kasonglee: @camilamacedo86 i will enhance the coverage in the new commit. Do i need to make a new pull-request?

leader/leader.go
```diff
@@ -182,6 +182,14 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 					if err != nil {
 						log.Error(err, "Leader pod could not be deleted.")
 					}
+				case isNotReadyNode(ctx, config.Client, leaderPod.Spec.NodeName):
+					log.Info("the status of the node where operator pod with leader lock was running has been 'notReady'")
+					log.Info("Deleting the leader.")
+
+					//Mark the termainating status to the leaderPod and Delete the configmap lock
```
jmrodri, line 189: Nit: should have a space after `//` : `// Mark `
Nit: typo `termainating` should be `terminating`.

leader/leader.go
```diff
@@ -270,3 +278,44 @@ func getPod(ctx context.Context, client crclient.Client, ns string) (*corev1.Pod
...
+	}
+	return false
+
+}
+
+func deleteLeader(ctx context.Context, client crclient.Client, leaderPod *corev1.Pod, existing *corev1.ConfigMap) error {
+	err := client.Delete(ctx, leaderPod)
+	if err != nil {
+		log.Error(err, "Leader pod could not be deleted.")
+		return err
```
jmrodri, line 310: This is all good. Pretty clean and easy to read.

leader/leader_test.go
jmrodri: @kasonglee I see you already added tests that is just fine. There is no need for a new commit.

jmrodri (approved):
two minor issue but I will not hold this PR up for it. I can fix those later. So far this looks fine to me.

Issue #27, opened 2020-08-21
Call golangci-lint via GitHub Action 


camilamacedo86:
## Feature Request

**Is your feature request related to a problem? Please describe.**
It allows we to check the lint issues in the PR which make the review and the process to contribute to it easier. 
See; https://github.com/golangci/golangci-lint-action

**Describe the solution you'd like**
Call golangci-lint via GitHub Action

PR #28, opened 2020-08-24
dev(dx): lint improvements

camilamacedo86:
**Description**
- Add makefile targets to run the lint --fix
- Add GitHub Action to call it and do the CI integration. 
- Ensure a standard across the projects

**Motivation**
- A better way to check to linting the project in the CI which allows automatically comments be done (pre-code review)

screen-shot-2020-08-21-at-17 19 57

Closes: https://github.com/operator-framework/operator-lib/issues/27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,13 +37,6 @@ jobs:
     - name: Build
       run: make build
 
-    - name: Lint
-      uses: golangci/golangci-lint-action@v1
-      with:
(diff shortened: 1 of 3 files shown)


joelanford:
@camilamacedo86 Any reason we moved the lint step into an entirely separate workflow rather than just a separate job in the existing CI workflow?

The workflow defines when to run the jobs, so it seems more maintainable to maintain the workflow's `on:` configuration once if we intend for the linter to always run during CI.

PR #33, opened 2020-08-28
.github/workflows: remove make build task (already covered in prow test)

joelanford:
Remove `make build` from github actions CI workflow.

All required tests in this repo are executed via prow.

The remaining GitHub actions are now optional checks that produce nice GH integrations with GolangCI lint and Coveralls. The `make build` step in the GH action is duplicative of a step in the prow tests and it doesn't produce any helpful integrations, so remove it to keep the GH workflow limited to the things that prow can't do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,9 +34,6 @@ jobs:
     - name: Get dependencies
       run: go mod download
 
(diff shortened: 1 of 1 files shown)


camilamacedo86 (approved):
/lgtm

PR #35, opened 2020-10-06
Remove 'status' package

varshaprasad96:
Remove status package which has helper
functions for Operator Conditions, in favor of
upstream conditions library.

The helpers are already present in [upstream](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go) library.

PS: Docs related to this change will be updated in SDK(https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#manage-cr-status-conditions).


jmrodri (approved):
/lgtm

PR #36, opened 2020-10-13
Add release script

varshaprasad96:
Add release script and corresponding steps.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+###############################################################################################
+### RELEASING OPERATOR-LIB
+# Every operator-lib release should have a corresponding git semantic version tag
+# begining with `v`, ex.n `v1.2.3`.
+#
+# STEP 1: Create a release branch with the name vX.Y.x. Example: git checkout -b v1.2.x
+#
+# STEP 2: Run the release script by providing the operator-lib release version as an argument
+# in the above mentioned format. Example: ./release vX.Y.Z
+#
+# STEP 3: This script will create a release tag locally. Push the release branch and tag:
+# git push upstream <release-branch>
+# git push upstream <tag-name>, wherein <tag-name> is the release version.
+#
+# STEP 4: Update the release notes in github with the changes included in corresponding
+# operator-lib version.
+#################################################################################################
+
+set -eu
+
+if [[ $# != 1 ]]; then
+	echo "usage: $0 vX.Y.Z"
+	exit 1
+fi
+
+VER=$1
+NUMRE="0|[1-9][0-9]*"
+PRERE="\-(alpha|beta|rc)\.[1-9][0-9]*"
+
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -0,0 +1,46 @@
...
+	exit 1
+fi
+
+VER=$1
+NUMRE="0|[1-9][0-9]*"
+
+if ! [[ "$VER" =~ ^v($NUMRE)\.($NUMRE)\.($NUMRE)?$ ]]; then
+	echo "malformed version: \"$VER\""
+	exit 1
+fi
```
jmrodri, line 34: I would actually allow the pre-release stuff we did for SDK too. We didn't need it until we did, then we had to figure out a way to allow it. So I'd reuse what we had over there:
https://github.com/operator-framework/operator-sdk/blob/master/release.sh#L10-L18

release.sh
```diff
@@ -0,0 +1,46 @@
...
+	echo "directory has uncommitted files"
+	exit 1
+fi
+
+# Run tests
+make check
+
+# Tag the release commit and verify its tag
+git tag --sign --message "operator-lib $VER" "$VER"
+git verify-tag --verbose $VER
```
jmrodri, line 46: I think we should offer the suggested next steps to the user:
```
git push upstream v0.1.x
git push upstream --tags
```
Or just print what we have in STEP 3 & 4 from the comment at the top as reminders to the callers there are 2 other steps left. 

```bash
echo "Don't forget to:
echo ""
echo "git push upstream <release-branch>"
echo "git push upstream $VER"
echo ""
echo "Then create a github release for the tag.
```

jmrodri (changes requested):
Just a minor addition to print out the next steps as reminders. The tool we had to build RPMs would do something like this:

```
[jesusr@camaro go-operator{master}]$ tito tag
Creating output directory: /tmp/tito
Tagging new version of go-operator: untagged -> 0.1.1-120201012224555.git.1.cab8ae7
Created tag: go-operator-0.1.1-120201012224555.git.1.cab8ae7
   View: git show HEAD
   Undo: tito tag -u
   Push: git push --follow-tags origin
```
It's always a good reminder that running `release.sh` isn't enough, and I don't have to go lookup another doc.

release.sh
varshaprasad96: Agreed! This is a good idea to add reminder !

release.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Add reminder on next stpes
+echo ""
+echo "Done forget to:"
+echo ""
+echo "git push upstream <release-branch>"
+echo "git push upstream $VER"
+echo ""
+echo "Also update the release notes in github for this tag."
```
joelanford, line 58: Not for this PR, but in the future it would be interesting to use [goreleaser](https://goreleaser.com/) to automatically produce the GitHub release when a tag is pushed.

https://goreleaser.com/cookbooks/release-a-library/

PR #40, opened 2020-11-19
Bump k8s dependencies and controller-runtime

varshaprasad96:
Bump k8s dependencirs to v1.19+ and controller-runtime to v0.7.0-alpha.6

k8s 1.19+ is needed to have helper functions for operator conditions
Ref:  [metav1.Conditions](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go)


handler/instrumented_enqueue_object.go
```diff
@@ -41,21 +41,21 @@ type InstrumentedEnqueueRequestForObject struct {
 
 // Create implements EventHandler, and creates the metrics.
 func (h InstrumentedEnqueueRequestForObject) Create(e event.CreateEvent, q workqueue.RateLimitingInterface) {
-	setResourceMetric(e.Meta, e.Object)
+	setResourceMetric(e.Object, e.Object)
```
estroz, line 44: Change these function's signatures to only take one object if possible
```suggestion
	setResourceMetric(e.Object)
```

jmrodri (approved):
/lgtm

PR #41, opened 2020-11-19
Add helpers for operator-conditions

varshaprasad96:
This PR adds helpers for the [Operator Conditions](https://github.com/operator-framework/api/blob/master/crds/operators.coreos.com_operatorconditions.yaml) custom resource introduced by OLM. With the help of this package users can:
1. set a specific condition in the CRD.
2. find or delete a specific condition in the CRD.
3. check the status of the specific condition in the CRD.

Provide support for working with Operator Conditions CRD.

### DIFF
--- a/conditions/operator_conditions.go
+++ b/conditions/operator_conditions.go
@@ -15,8 +15,11 @@
 package conditions
 
 import (
+	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"strings"
 
 	api "github.com/operator-framework/api/pkg/operators/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -36,16 +39,29 @@ var (
 // which is set for the Condition resource owned by the operator.
 const operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
 
+var readNamespace = func() ([]byte, error) {
+	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+}
+
+// GetNamespacedName returns the NamespacedName of the CR. It returns an error
+// when the name of the CR cannot be found from the environment variable set by
+// OLM, or when the namespace cannot be found from the associated service account
+// secret.
 func GetNamespacedName() (*types.NamespacedName, error) {
 	conditionName := os.Getenv(operatorCondEnvVar)
 	if conditionName == "" {
 		return nil, fmt.Errorf("required env %s not set, cannot find the operator condition for the operator", operatorCondEnvVar)
 	}
-	// TODO
-	return nil, nil
+	operatorNs, err := getOperatorNamespace()
+	if err != nil {
+		return nil, err
+	}
+	return &types.NamespacedName{Name: conditionName, Namespace: operatorNs}, nil
 }
 
-func SetConditionStatus(operatorCondition *api.OperatorCondition, newCond metav1.Condition) error {
+// SetOperatorCondition adds the specific condition to the Condition CR or
+// updates the provided status of the condition if already present.
+func SetOperatorCondition(operatorCondition *api.OperatorCondition, newCond metav1.Condition) error {
 	if operatorCondition == nil {
 		return ErrNoOperatorCondition
 	}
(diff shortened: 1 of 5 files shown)


conditions/operator_conditions_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+			val, err := IsConditionStatusPresentAndEqual(operatorCondition, conditionType, conditionStatus)
+			Expect(err).To(HaveOccurred())
+			Expect(val).To(BeFalse())
+		})
+
+	})
+
+	Describe("GetNamespacedName", func() {
+		It("should error when name of the operator condition cannot be found", func() {
```
jmrodri, line 290: The only nit I have about the test is the order doesn't match the order of the methods in the operator_conditions.go file, so it was hard to review since I had to search for the tests for each method. I won't hold this PR up for this but just in general it makes it easier to review.

conditions/operator_conditions_test.go
```diff
@@ -0,0 +1,338 @@
...
+			err := SetOperatorCondition(nil, newCond)
+			Expect(err).To(HaveOccurred())
+			Expect(err).Should(MatchError(ErrNoOperatorCondition))
+		})
+	})
+
+	Describe("RemoveOperatorCondition", func() {
+		var (
+			rmvConditionType string
+		)
```
jmrodri, line 115: Nit: I would've just assigned it directly in the test for each one. 

```golang
rmvConditionType := conditionFoo
```

jmrodri (approved):
/lgtm 
Really great stuff, thorough testing, easy to read methods. Love it.

conditions/operator_conditions.go
```diff
@@ -0,0 +1,133 @@
...
+// GetNamespacedName returns the NamespacedName of the CR. It returns an error
+// when the name of the CR cannot be found from the environment variable set by
+// OLM, or when the namespace cannot be found from the associated service account
+// secret.
+func GetNamespacedName() (*types.NamespacedName, error) {
+	conditionName := os.Getenv(operatorCondEnvVar)
+	if conditionName == "" {
+		return nil, fmt.Errorf("required env %s not set, cannot find operator condition CR for the operator", operatorCondEnvVar)
+	}
+	operatorNs, err := getOperatorNamespace()
```
estroz, line 52: This code is duplicated in the [leader](https://github.com/operator-framework/operator-lib/blob/main/leader/leader.go#L240) package. Can you pull this out into a top-level internal package?

Additionally Operators that use this code won’t be able to run locally unless the environment creates the SA namespace file on each run. Unlike the leader package, which shouldn’t be executed, conditions should be set when running locally, so there needs to be a way to set this namespace. I suggest an environmental variable like `OPERATOR_NAMESPACE_LOCAL` that gets read when `getOperatorNamespace()` returns `ErrNoNamespace`

conditions/operator_conditions.go
```diff
@@ -0,0 +1,133 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var (
+	// ErrNoOperatorCondition indicates that the operator condition CRD is nil
+	ErrNoOperatorCondition = fmt.Errorf("operator Condition CRD is nil")
+)
+
+// TODO: verify from OLM if this will be the name of the environment variable
```
estroz, line 35: @awgreene can you confirm? Afaik this is the name discussed in the EP

conditions/operator_conditions.go
```diff
@@ -0,0 +1,133 @@
...
+// which is set for the Condition resource owned by the operator.
+const operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
```
estroz, line 37: This needs godocs

conditions/operator_conditions.go
awgreene: This name is fine, if you prefer something else just let me know.

varshaprasad96:
@estroz @jmrodri : addressed review comments in #42 . Please have a look.

PR #42, opened 2020-11-21
Address review comments for operator conditions helpers

varshaprasad96:
This PR does the following:

* Move getOperatorNamespace() to a seprate utils package.
* Allow namespace to be set locally with an env variable.
* Modify the order of tests and add godoc to env variables.
* Remove redundant code.

Follow-up #41

### DIFF
--- a/conditions/operator_conditions.go
+++ b/conditions/operator_conditions.go
@@ -16,7 +16,6 @@ package conditions
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 
 	api "github.com/operator-framework/api/pkg/operators/v1"
@@ -38,9 +37,7 @@ const (
 	operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
 )
 
-var readNamespace = func() ([]byte, error) {
-	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
-}
+var readNamespace = utils.GetOperatorNamespace
 
 // GetNamespacedName returns the NamespacedName of the CR. It returns an error
 // when the name of the CR cannot be found from the environment variable set by
@@ -53,7 +50,7 @@ func GetNamespacedName() (*types.NamespacedName, error) {
 	if conditionName == "" {
 		return nil, fmt.Errorf("could not determine operator condition name: environment variable %s not set", operatorCondEnvVar)
 	}
-	operatorNs, err := utils.GetOperatorNamespace(readNamespace)
+	operatorNs, err := readNamespace()
 	if err != nil {
 		return nil, fmt.Errorf("could not determine operator namespace: %v", err)
 	}
--- a/conditions/operator_conditions.go
+++ b/conditions/operator_conditions.go
@@ -15,13 +15,12 @@
 package conditions
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
-	"strings"
 
 	api "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/operator-lib/internal/utils"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -32,9 +31,12 @@ var (
 	ErrNoOperatorCondition = fmt.Errorf("operator Condition CRD is nil")
 )
 
-// TODO: verify from OLM if this will be the name of the environment variable
-// which is set for the Condition resource owned by the operator.
-const operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+const (
+	// operatorCondEnvVar is the env variable which
+	// contains the name of the Condition CR associated to the operator,
+	// set by OLM.
+	operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+)
 
 var readNamespace = func() ([]byte, error) {
 	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
@@ -42,16 +44,18 @@ var readNamespace = func() ([]byte, error) {
 
 // GetNamespacedName returns the NamespacedName of the CR. It returns an error
 // when the name of the CR cannot be found from the environment variable set by
-// OLM, or when the namespace cannot be found from the associated service account
-// secret.
+// OLM. Hence, GetNamespacedName() can provide the NamespacedName when the operator
+// is running on cluster and is being managed by OLM. If running locally, operator
+// writers are encouraged to skip this method or gracefully handle the errors by logging
+// a message.
 func GetNamespacedName() (*types.NamespacedName, error) {
 	conditionName := os.Getenv(operatorCondEnvVar)
 	if conditionName == "" {
-		return nil, fmt.Errorf("required env %s not set, cannot find operator condition CR for the operator", operatorCondEnvVar)
+		return nil, fmt.Errorf("could not determine operator condition name: environment variable %s not set", operatorCondEnvVar)
 	}
-	operatorNs, err := getOperatorNamespace()
+	operatorNs, err := utils.GetOperatorNamespace(readNamespace)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("could not determine operator namespace: %v", err)
 	}
 	return &types.NamespacedName{Name: conditionName, Namespace: operatorNs}, nil
 }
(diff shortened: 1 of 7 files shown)


conditions/operator_conditions.go
```diff
@@ -32,9 +32,15 @@ var (
...
-// which is set for the Condition resource owned by the operator.
-const operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+const (
+	// operatorCondEnvVar is the env variable which
+	// contains the name of the Condition CR associated to the operator,
+	// set by OLM.
+	operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+	// OperatorNsEnvVar is the env variable which can be used to set the namespace
+	// of the operator if running locally.
+	OperatorNsEnvVar = "OPERATOR_NAMESPACE_LOCAL"
```
jmrodri, line 42: Why is one exported and the other not?

leader/leader_test.go
```diff
@@ -70,17 +71,17 @@ var _ = Describe("Leader election", func() {
 		})
 		It("should return an ErrNoNamespace", func() {
 			os.Setenv("POD_NAME", "leader-test")
-			readNamespace = func() ([]byte, error) {
+			readNs = func() ([]byte, error) {
```
jmrodri, line 74: Why rename it?

jmrodri (commented):
Just a few questions

conditions/operator_conditions.go
varshaprasad96: `operatorCondEnvVar` will be used by OLM to set operator conditions name and `OperatorNsEnvVar` will be used by the users. But yeah, `OperatorNsEnvVar` need not be exported as its just an env variable. Modified it.

conditions/operator_conditions.go
```diff
@@ -32,9 +32,15 @@ var (
...
-// which is set for the Condition resource owned by the operator.
-const operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+const (
+	// operatorCondEnvVar is the env variable which
+	// contains the name of the Condition CR associated to the operator,
+	// set by OLM.
+	operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+	// operatorNsEnvVar is the env variable which can be used to set the namespace
+	// of the operator if running locally.
+	operatorNsEnvVar = "OPERATOR_NAMESPACE_LOCAL"
```
joelanford, line 42: If an operator is running locally, then it seems pretty unlikely that OLM will know anything about it, and if that's true, there won't be an OperatorCondition object anyway.

IMO, we should:
1. Have `GetNamespacedName()` return a checkable errors (i.e. errors that can be checked with `errors.Is()` or `errors.As()`) if it cannot find the namespace file or the operator condition environment variable.
2. Put good GoDoc comments on `GetNamespacedName()` that explains when it works (generally only in a cluster being managed by OLM), what errors callers should expect when it fails, and maybe an example of how to handle those errors.

If I we're writing an operator using this, I think I would call this function all the time, and then gracefully handle the errors by logging a message and/or setting some configuration value like `isOLMDeployed := false`, which could be used to disable the code that manages the `OperatorCondition`.

How about something like:

```go
type ErrGetNamespacedName struct {
	Error error
}

func (e ErrGetNamespacedName) Error() string {
	return fmt.Sprintf("could not determine operator condition namespaced name: %v", e.Error)
}

func GetNamespacedName() {
	conditionName := os.Getenv(operatorCondEnvVar)
	if conditionName == "" {
		return nil, &ErrGetNamespacedName{
			Error: fmt.Errorf("environment variable %s not set", operatorCondEnvVar),
		}
	}
	operatorNs, err := getOperatorNamespace()
	if err != nil {
		return nil, &ErrGetNamespacedName{
			Error: fmt.Errorf("could not determine operator namespace: %v", err),
		}
	}
	return &types.NamespacedName{Name: conditionName, Namespace: operatorNs}, nil
}
```

leader/leader.go
```diff
@@ -31,10 +31,6 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// ErrNoNamespace indicates that a namespace could not be found for the current
-// environment
-var ErrNoNamespace = fmt.Errorf("namespace not found for current environment")
-
```
joelanford, line 37: Moving this error to an `internal` directory is a breaking change because callers can no longer access it in their error checks.

IMO, we should try to keep libraries somewhat self-contained so that users can call `leader.Become()` and check any errors without having to import a separate package to check for errors it might return.

conditions/operator_conditions.go
```diff
@@ -119,15 +125,21 @@ func IsConditionStatusPresentAndEqual(operatorCondition *api.OperatorCondition,
...
+			if ns == "" {
+				return "", fmt.Errorf(" %s: environment variable %s not set", errMsg, operatorCondEnvVar)
+			}
+			return ns, nil
 		}
-		return "", err
+		return "", fmt.Errorf("%s: %v", errMsg, err)
 	}
-	ns := strings.TrimSpace(string(nsBytes))
 	return ns, nil
```
joelanford, line 144: Based on my comment above, remove this function and call `utils.GetOperatorNamespace()` directly?

conditions/operator_conditions.go
joelanford: cc @estroz

joelanford (commented):
Not to throw a wrench into things, but another potential way to go with this library is to make an interface like this:

```go
type Upgradabler interface {
    // GetUpgradable gets the Upgradable condition on the operator's 
	// OperatorCondition. It returns an error if there are problems getting
	// the OperatorCondition object.
    GetUpgradable(context.Context) (*metav1.Condition, error)

    // SetUpgradable sets the Upgradable condition on the operator's 
	// OperatorCondition to the status s. It returns an error if there
	// are problems getting or updating the OperatorCondition object.
    SetUpgradable(context.Context, metav1.ConditionStatus) error
}
```

And then we could have an implementation that internally looks up (and caches) the namespaced name, and then fetches the current OperatorCondition, sets (or gets) the condition value, and then performs an update (if the value changed).

This way, a user wouldn't have to import `operator-framework/api` or really even know about the `OperatorCondition` type, and it would make it much easier for operator authors to mock an implementation of this and write tests around their upgrade logic.

conditions/operator_conditions.go
varshaprasad96: This would mean that `getOperatorNamespace` is present in both the packages, instead of the utils package. Declaring the error twice and checking for the error from utils.GetOperatorNamespace() to be equal to to the one defined in `leader.go` would not work.

leader/leader.go
estroz: @varshaprasad96 this needs to be re-added. You could do something like
```go
var ErrNoNamespace = utils.ErrNoNamespace
```

varshaprasad96:
@estroz @jmrodri : Made the suggested changes, please have a look

internal/utils/utils.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"strings"
+)
+
+// ErrNoNamespace indicates that a namespace could not be found for the current
+// environment
+var ErrNoNamespace = fmt.Errorf("namespace not found for current environment")
+
+// GetOperatorNamespace returns the namespace the operator should be running in.
+func GetOperatorNamespace(readNamespace func() ([]byte, error)) (string, error) {
```
estroz, line 28: Alternative approach:
Instead of doing
```go
func GetOperatorNamespace(readNamespace func() ([]byte, error)) (string, error) {
	...
}

var readNamespace = func() ([]byte, error) {
	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
}

func foo() {
	_, _ = GetOperatorNamespace(readNamespace)
}
```

you could do

```go
func GetOperatorNamespace() (string, error) {
	...
}

var readNamespace = GetOperatorNamespace()

func foo() {
	_, _ = readNamespace()
}
```

which is a little less confusing and still mockable in tests in a similar way as to how you have it now:

```go
var _ = Describe("test namespace", func() {
	// This
	readNamespace = func() (string, error) {
		return "some-ns"
	}
	// Or this
	f := func(s string) (func() (string, error)) {
		return func() (string, error) { return s, nil }
	}
	readNamespace = f("some-ns")
}
```

internal/utils/utils.go
```diff
@@ -0,0 +1,44 @@
...
+// environment
+var ErrNoNamespace = fmt.Errorf("namespace not found for current environment")
+
+var readSAFile = func() ([]byte, error) {
+	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+}
+
+// GetOperatorNamespace returns the namespace the operator should be running in from
+// the associated service account secret.
+var GetOperatorNamespace = func() (string, error) {
```
varshaprasad96, line 34: Since go would not allow `var readNamespace = GetOperatorNamespace()` when `GetOperatorNamespace()` is not an anonymous function, have declared it as a variable here. We would still need `readSAFile` to test this function.

estroz (approved):
/lgtm

PR #43, opened 2020-12-04
Add library to interact with operator conditions

varshaprasad96:
This PR removes the previous operator_conditions package and
introduces a `Conditions` interface, using which one can:
1. Get an existing condition from the operator conditions CR in cluster.
2. Create/Update a condition in the operator condition CR.

EP: [Operator-Conditions lib](https://github.com/operator-framework/enhancements/blob/master/enhancements/operator-conditions-lib.md)


jmrodri (approved):
/lgtm
This looks a lot better. Great work!

testbin/setup-envtest.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
```
estroz, line 1: Nit: this file should be ignored by adding `**/testbin/*` to the `.gitignore`.

PR #44, opened 2020-12-04
Use fakeclient instead of envtest

jmrodri:
Remove the reliance on envtest for testing this library. Switched to using just the controller-runtime fakeclient.

Keep the dependencies to a minimum


estroz (approved):
This is probably ok because these are unit tests and are not testing complex behavior.

/approve
/lgtm

PR #51, opened 2021-01-24
Add option to read the operator namespace from an envirnment variable…

### LINKED ISSUE #50: Can't run operators locally, when using the conditions package
## Bug Report

The `NewCondition` (and other) method, called (indirectlly) to the `readSAFile` method. This method uses a hard code path for the operator namespace file:
https://github.com/operator-framework/operator-lib/blob/c0ba7dcbdce825f752802c871a0b52a37bfbf135/internal/utils/utils.go#L29

This implementation blocks the abilit to run the operator locally for develpment and debug, 

Please consider one of the following, or both (pefered):
1. Replace the hard-code path with a environment variable, with a default of the current value.
2. Use an environment variable to store the namespace, and use the file only if this variable is not set.

**Environment**
* operator-lib version:

nunnatsa:
…, or from a non-standard SA file.

While developing an operator, it is impossible to run from a local environment variable if the code
is trying to access the namespace file. For example, calling the `conditions.NewCondition` function
will return an error when the operator is running in a development environment.

This PR add to options to override the hard-coded path to the namespace file with another location:
1. by setting the new `OPERATOR_NAMESPACE` environment variable to the required namespace.
2. by setting the new `SA_FILE_PATH` environment variable to the local path of the namespace file.

If both new environment variables are set, the namespace will be taken from the `OPERATOR_NAMESPACE`
environment variable.

Fixes #50

Signed-off-by: Nahshon Unna-Tsameret <nunnatsa@redhat.com>

### DIFF
--- a/internal/utils/testfiles/namespace
+++ b/internal/utils/testfiles/namespace
@@ -1,0 +1,1 @@
+testnamespace
--- a/internal/utils/testfiles/namespaceWithSpaces
+++ b/internal/utils/testfiles/namespaceWithSpaces
@@ -1,0 +1,3 @@
+  testnamespace
+
+
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -21,17 +21,41 @@ import (
 	"strings"
 )
 
+const (
+	// SAFileDefaultLocation default location of the service account namespace file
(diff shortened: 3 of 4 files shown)


estroz:
Closing, see https://github.com/operator-framework/operator-lib/issues/50#issuecomment-767122772

/close

PR #56, opened 2021-03-31
patterns/singleton: add Singleton pattern helpers

estroz:
**Description of the change:** This PR adds utilities for the singleton pattern, which constrains the size of the set of all objects of some type to 1, with a pre-defined name. `NewRunnable()` will create singleton objects at manager startup time (optional). `ValidateCreate()` will fail if an object being created does not have the expected object name, and `ValidateDelete()` will log an error when the singleton object is deleted.

**Motivation for the change:** this lib should host pattern utilities to spread best practices.

/kind feature


joelanford (commented):
I'm curious if it makes sense to enforce the singleton constraint via an admission webhook rather than a reconciler that deletes objects (probably right after they're created).

From the perspective of the user creating the custom object that violates the constraint, it would be _much_ better to get an error message from the API server that their object can't be created due to a singleton constraint rather than a mysterious disappearance of the thing they just created.

Related: it would be pretty kinda cool to make a singleton plugin for the `create api` and/or `create webhook` commands that scaffolds everything needed for a singleton api/controller/webhook.

patterns/singleton/singleton.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (r runnable) NeedLeaderElection() bool { return true }
+
+// TODO(estroz): parallelize
+func (r runnable) Start(ctx context.Context) error {
+	for _, obj := range r.objs {
+		if err := r.c.Create(ctx, obj); err != nil {
+			return err
+		}
```
joelanford, line 72: What happens if these objects are deleted? Since the operator creates them, it seems it would be in-scope for it to re-create them and patch them back to some desired state. Perhaps a `SingletonReconciler` that makes sure the desired objects are always present.

estroz:
The webhook approach does make more sense, and I think this has been discussed in the past. I’ll change this to implement webhook helpers.

patterns/singleton/singleton.go
estroz: @joelanford I'm thinking that [`controllerutil.CreateOrUpdate()`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil?utm_source=godoc#CreateOrUpdate) should be used directly rather than writing a reconciler here, as the reconciler would basically just be calling this function.

jmrodri (approved):
/lgtm

PR #58, opened 2021-04-20
Add conditions.Factory

### LINKED ISSUE #50: Can't run operators locally, when using the conditions package

joelanford:
Add conditions.Factory

The conditions.NewCondition function reads cluster state to determine
the namespace the operator is running in so that it can manipulate the
correct OperatorConditions object in a cluster.

However this function's dependency on a specific file existing in a
specific location makes it difficult to test code that uses it.

The Factory struct is introduced to enable users to provide their own
implementations for looking up the name and namespace of the operator's
OperatorCondition object. Users can now build a factory and inject it into
code that builds new conditions. This enables code under test to use the
Factory abstraction and enables operator authors to inject custom
functionality needed when writing tests.

Closes #50


conditions/factory.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const (
+	// operatorCondEnvVar is the env variable which
+	// contains the name of the Condition CR associated to the operator,
+	// set by OLM.
+	operatorCondEnvVar = "OPERATOR_CONDITION_NAME"
+)
+
+// DefaultGetConditionName reads and returns the OPERATOR_CONDITION_NAME environment
```
estroz, line 67:
```suggestion
// getConditionName reads and returns the OPERATOR_CONDITION_NAME environment
```

conditions/factory.go
```diff
@@ -0,0 +1,94 @@
...
+	if name == "" {
+		return "", fmt.Errorf("could not determine operator condition name: environment variable %s not set", operatorCondEnvVar)
+	}
+	return name, nil
+}
+
+// readNamespace gets the namespacedName of the operator.
+var readNamespace = utils.GetOperatorNamespace
+
+// DefaultGetConditionNamespace reads the namespace file mounted into a pod in a
```
estroz, line 80:
```suggestion
// getConditionNamespace reads the namespace file mounted into a pod in a
```

joelanford:
Closing this since there doesn't seem to be much interest in it. We can always re-visit/re-open if this comes up again.

nunnatsa:
Hi @joelanford. We want to re-try implementing the upgradable condition in [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator), now when [bugzilla bug #1927340](https://bugzilla.redhat.com/show_bug.cgi?id=1927340) is fixed. So we'll need a fix for #50.

priyanka19-98:
We want to implement upgradeable condition in [OCS](https://github.com/openshift/ocs-operator) , and will need  fix for #50

nunnatsa:
@joelanford - could you please reconsider this? It will really help to have a solution for #50, to get more confidence in our implementation, because this is not easy to verify the behavior of the operator regarding the operator condition in e2e tests.

nunnatsa:
/lgtm

@varshaprasad96 - could you please merge this?

PR #59, opened 2021-04-29
Ignore time in ManagedFields when processing events from dependent resources

fabianvf:
Ignores time in ManagedFields when determining whether events should be skipped. This gets us back the behavior we had before managedFields were added, as when the kubelet updates the resourceVersion it also updates the time in its managedFields entry.

We've had users reporting infinite reconciliation and this is likely the culprit.

Fixes https://github.com/operator-framework/operator-sdk/issues/3950
Fixes https://github.com/operator-framework/operator-sdk/issues/4258

### DIFF
--- a/predicate/dependent.go
+++ b/predicate/dependent.go
@@ -17,6 +17,7 @@ package predicate
 import (
 	"reflect"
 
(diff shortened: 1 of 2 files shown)


jmrodri (approved):
/lgtm

PR #60, opened 2021-05-25
handler,predicate: pause object reconciliation by annotation

estroz:
**Description of the change:** added `NewPause(key)` to handler and predicate packages that returns an EventHandler and Predicate, respectively, to pause object reconciliation on objects with the `<key>: true` annotation.

**Motivation for the change:** something along the lines of https://github.com/operator-framework/operator-sdk/issues/3418.

/kind feature

### DIFF
--- a/internal/annotation/filter.go
+++ b/internal/annotation/filter.go
@@ -15,7 +15,7 @@
 // Package annotation contains event handler and predicate builders for annotations.
 // There are two types of builders:
 //
-// - Falsy builders result in objects not being queued if the annotation is not present OR contains a falsy value.
+// - Falsy builders result in objects being queued if the annotation is not present OR contains a falsy value.
 // - Truthy builders are the falsy complement: objects will be enqueued if the annotation is present AND contains a truthy value.
 //
 // Truthiness/falsiness is determined by Go's strconv.ParseBool().
@@ -51,7 +51,7 @@ func NewFalsyPredicate(key string, opts Options) (predicate.Predicate, error) {
 	return newFilter(key, opts)
 }
 
-// NewFalsyEventHandler returns an event handler that adds objects to a queue
+// NewFalsyEventHandler returns an event handler that enqueues objects
 // that do not have annotation with key string key or whose value is falsy.
 func NewFalsyEventHandler(key string, opts Options) (handler.EventHandler, error) {
 	opts.truthy = false
@@ -65,7 +65,7 @@ func NewTruthyPredicate(key string, opts Options) (predicate.Predicate, error) {
 	return newFilter(key, opts)
 }
 
-// NewTruthyEventHandler returns an event handler that adds objects to a queue
+// NewTruthyEventHandler returns an event handler that enqueues objects
 // that do have annotation with key string key and whose value is truthy.
 func NewTruthyEventHandler(key string, opts Options) (handler.EventHandler, error) {
 	opts.truthy = true
--- a/internal/annotation/filter.go
+++ b/internal/annotation/filter.go
@@ -1,0 +1,210 @@
+// Copyright 2021 The Operator-SDK Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package annotation contains event handler and predicate builders for annotations.
+// There are two types of builders:
+//
+// - Falsy builders result in objects not being queued if the annotation is not present OR contains a falsy value.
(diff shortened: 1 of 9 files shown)


estroz:
The tests added here break other handler tests because they share the same metrics registry. If this POC is accepted, the author (me or someone else) will re-write the tests so they do not affect each other.

internal/annotation/pause.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package annotation
+
+// PausedKey being present on an object with a truthy value
+// results in that object not passing a predicate or being added to a queue.
+// Read: this maps to a falsy filter (think truthy complement).
+const PausedKey = "operatorframework.io/paused"
```
fabianvf, line 20: Out of curiousity, for non-sdk namespaced annotations (which I assume would be like `sdk.operatorframework.io` or something), is there any place where these more global operatorframework annotations would be tracked? 100% not a blocker. just curious if there's any risk that another project like OLM might want to add something similar but perhaps with slightly different semantics and not realize that we already have claimed this annotation?

estroz: There is none currently. I'd expect that would be documented in the api repo under some `docs/annotations/<repo or project or component>.md` file. Great call-out.

joelanford: Alternatively, we could just make the annotation key an input into the library to skirt this problem.

From an operator author perspective, I would rather not leak implementation details (like that I'm using operatorframework) to end users. Instead, I could use `my.domain/paused`.

estroz: https://github.com/operator-framework/api/issues/122

estroz: >From an operator author perspective, I would rather not leak implementation details (like that I'm using operatorframework) to end users. Instead, I could use my.domain/paused

Right I do see this as a concern, but does having non-standard namespacing break [annotation conventions](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set)? Also is it really that much more of a leap to infer that if `my.domain/paused` is set that a controller is using operator-lib's pause mechanism? The relationship becomes implicit but is still there via this repo.

estroz: >We'd be delegating annotation convention adherence to the operator author, right? Or we could potentially validate that the provided string follows the convention.

I'm more concerned with semantics rather than format, from the angle of "operator-framework owns the pause mechanism, and so the annotation should be namespaced as such". However I am amenable to the argument that just because you're using a library does not mean you must delegate ownership of some process to that library (unless ownership delegation is required by library contract, which it does not have to be here).

>On the flipside, if an end user is able to determine that a project is using operator framework, will they assume that they can use this annotation?

I think this is a potential security issue actually. What if a user can pause their CR themselves? This annotation should be applicable by admins only.

>end users should go to operator docs (not operatorframework docs) to determine what annotations are supported by an operator and what each of those annotations does.

In general I agree, for non-admin annotations. But for admin-only annotations like this pause feature, the meaning of an annotation should be opaque and no docs should exist for end users.

joelanford: > I think this is a potential security issue actually. What if a user can pause their CR themselves? This annotation should be applicable by admins only.

The `metadata.annotations` field is not separable from the rest of the custom object from an RBAC perspective, so the only way to restrict a particular annotation from being set would be to introduce an admission webhook, right?

estroz: Yeah, so this is out of scope here but should be noted in the docs. Is there a typical way one would write an admission controller to check which RBAC subject is modifying an object?

internal/annotation/filter.go
```diff
@@ -0,0 +1,210 @@
...
+	annotation := map[string]string{key: fmt.Sprintf("%v", truthy)}
+	return validation.ValidateAnnotations(annotation, fldPath).ToAggregate()
+}
+
+// filter implements a filter for objects with a truthy "paused" annotation (see Key).
+// When this annotation is removed or value does not evaluate to "true",
+// the controller will see events from these objects again.
+type filter struct {
+	key  string
+	ret  bool
```
joelanford, line 142: Non-blocking thought:

It might be generally useful to add a `Not(in predicate.Predicate) predicate.Predicate` wrapper in C-R that inverts the logic of the input predicate.

If that existed, would you need the `ret` field?

I alluded to this in the C-R PR where I added the `Or` and `And` wrappers [here](https://github.com/kubernetes-sigs/controller-runtime/pull/1043#issuecomment-656872379), but I never added it because there hasn't been a use case for it.

internal/annotation/pause.go
joelanford: The subject is included in the admission request object. Docs here: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request

estroz:
@joelanford given that Falsy/Truthy filters are complements, yes `Not()` would be useful here.

internal/annotation/filter.go
estroz: I don't think you'd need `ret` here in that case. Given that operator-lib is pre v1.0, are you ok with adding this as-is now then removing if `Not` is added to c-r?

jmrodri (approved):
/lgtm

estroz:
Not sure why it's not marking this automatically approved. @jmrodri can you `/approve` too?

varshaprasad96 (approved):

PR #61, opened 2021-05-26
Update condition lib to use v2

awgreene:
The v2 OperatorCondition CRD was introduced to the operator-framework/api
repository. This change updates the condition library to consume the new
changes.

### DIFF
--- a/conditions/conditions.go
+++ b/conditions/conditions.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"os"
 
-	apiv1 "github.com/operator-framework/api/pkg/operators/v1"
+	apiv2 "github.com/operator-framework/api/pkg/operators/v2"
 	"github.com/operator-framework/operator-lib/internal/utils"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -46,7 +46,7 @@ const (
 // conditionType in the OperatorCondition CR.
 type condition struct {
 	namespacedName types.NamespacedName
-	condType       apiv1.ConditionType
+	condType       apiv2.ConditionType
 	client         client.Client
 }
 
@@ -55,7 +55,7 @@ var _ Condition = &condition{}
 // NewCondition returns a new Condition interface using the provided client
 // for the specified conditionType. The condition will internally fetch the namespacedName
 // of the operatorConditionCRD.
-func NewCondition(cl client.Client, condType apiv1.ConditionType) (Condition, error) {
+func NewCondition(cl client.Client, condType apiv2.ConditionType) (Condition, error) {
 	objKey, err := GetNamespacedName()
 	if err != nil {
 		return nil, err
@@ -69,12 +69,12 @@ func NewCondition(cl client.Client, condType apiv1.ConditionType) (Condition, er
 
 // Get implements conditions.Get
 func (c *condition) Get(ctx context.Context) (*metav1.Condition, error) {
-	operatorCond := &apiv1.OperatorCondition{}
+	operatorCond := &apiv2.OperatorCondition{}
 	err := c.client.Get(ctx, c.namespacedName, operatorCond)
 	if err != nil {
 		return nil, ErrNoOperatorCondition
 	}
-	con := meta.FindStatusCondition(operatorCond.Status.Conditions, string(c.condType))
+	con := meta.FindStatusCondition(operatorCond.Spec.Conditions, string(c.condType))
 
 	if con == nil {
 		return nil, fmt.Errorf("conditionType %v not found", c.condType)
@@ -84,7 +84,7 @@ func (c *condition) Get(ctx context.Context) (*metav1.Condition, error) {
 
 // Set implements conditions.Set
 func (c *condition) Set(ctx context.Context, status metav1.ConditionStatus, option ...Option) error {
-	operatorCond := &apiv1.OperatorCondition{}
+	operatorCond := &apiv2.OperatorCondition{}
 	err := c.client.Get(ctx, c.namespacedName, operatorCond)
 	if err != nil {
 		return ErrNoOperatorCondition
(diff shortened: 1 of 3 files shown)


varshaprasad96 (commented):
A quick doubt, is OLM still going to support v1 condition CRD's for a release and then deprecate (eventually remove) later? If thats the case, I suppose this condition lib should also be versioned in a way to support both the condition CRD versions till OLM removes `v1 Conditions`. Or is it not necessary? 
cc: @estroz @jmrodri

awgreene:
> A quick doubt, is OLM still going to support v1 condition CRD's for a release and then deprecate (eventually remove) later? If thats the case, I suppose this condition lib should also be versioned in a way to support both the condition CRD versions till OLM removes `v1 Conditions`. Or is it not necessary?
> cc: @estroz @jmrodri

OperatorCondition V1 does not handle race conditions so it is broken. We do not want anyone using v1.

dinhxuanvu (approved):
/lgtm

estroz:
So clearly there is not enough test coverage _somewhere_ that resulted in this race condition not being caught. Has that coverage been implemented somewhere? Are more tests needed here?

conditions/conditions_test.go
```diff
@@ -75,17 +76,17 @@ var _ = Describe("Condition", func() {
...
 			Name:      "operator-condition-test",
 			Namespace: ns,
 		}
 
 		BeforeEach(func() {
-			operatorCond = &apiv1.OperatorCondition{
+			operatorCond = &apiv2.OperatorCondition{
 				ObjectMeta: metav1.ObjectMeta{Name: "operator-condition-test", Namespace: ns},
-				Status: apiv1.OperatorConditionStatus{
+				Spec: apiv2.OperatorConditionSpec{
```
varshaprasad96, line 89: a non-blocker, but it would be helpful to have [`conditionType`](https://github.com/operator-framework/api/blob/bb9b80e8278978efdb06a6f5d5682eb3cad330ec/pkg/operators/v1/operator_types.go#L19) declared in v2 too. Its odd that for someone using this library, they would have to use `conditionType` from `v1` and the OperatorCondition api from `v2`. Example - [here](https://github.com/operator-framework/operator-lib/blob/064486c30aa267578d6c552eb796ad99f969c213/conditions/conditions.go#L77).

If not the other option is - we can also create a follow up PR to remove the `conditionType` itself and let users specify a string directly.

awgreene:
>So clearly there is not enough test coverage somewhere that resulted in this race condition not being caught. Has that coverage been implemented somewhere? Are more tests needed here?

@estroz this should be handled in the OLM PR https://github.com/operator-framework/operator-lifecycle-manager/pull/2160/

conditions/conditions_test.go
dinhxuanvu: Addressed in this release: https://github.com/operator-framework/api/releases/tag/v0.9.1

varshaprasad96 (approved):

estroz:
Reposting from https://github.com/operator-framework/api/pull/119 for visibility:
> Out of curiosity: why v2 and not v2beta1? Are we confident no significant changes will be needed soon? I feel that it’s a bit hasty to just throw a new stable version out, especially given how short the lifetime of v1 was.

dinhxuanvu:
> Reposting from [operator-framework/api#119](https://github.com/operator-framework/api/pull/119) for visibility:
> 
> > Out of curiosity: why v2 and not v2beta1 [or even v2alpha1]? Are we confident no significant changes will be needed soon? I feel that it’s a bit hasty to just throw a new stable version out, especially given how short the lifetime of v1 was.

I don't think we will have another significant changes anytime soon. Any other changes are probably on OLM side, not on the CRD side but I'm fine with keeping it `v2beta1` if we all agree on that. cc @awgreene

estroz:
Resolved in https://github.com/operator-framework/api/pull/119#issuecomment-849794988. v2 it is.

awgreene:
@varshaprasad96 can we get eyes on this pr?

conditions/conditions.go
```diff
@@ -84,7 +84,7 @@ func (c *condition) Get(ctx context.Context) (*metav1.Condition, error) {
 
 // Set implements conditions.Set
 func (c *condition) Set(ctx context.Context, status metav1.ConditionStatus, option ...Option) error {
-	operatorCond := &apiv1.OperatorCondition{}
+	operatorCond := &apiv2.OperatorCondition{}
 	err := c.client.Get(ctx, c.namespacedName, operatorCond)
 	if err != nil {
 		return ErrNoOperatorCondition
```
nunnatsa, line 90: Please consider returning the original error, or at least wrap it.

conditions/conditions.go
```diff
@@ -69,12 +69,12 @@ func NewCondition(cl client.Client, condType apiv1.ConditionType) (Condition, er
 
 // Get implements conditions.Get
 func (c *condition) Get(ctx context.Context) (*metav1.Condition, error) {
-	operatorCond := &apiv1.OperatorCondition{}
+	operatorCond := &apiv2.OperatorCondition{}
 	err := c.client.Get(ctx, c.namespacedName, operatorCond)
 	if err != nil {
 		return nil, ErrNoOperatorCondition
```
nunnatsa, line 75: Please consider returning the original error, or at least wrap it.

conditions/conditions.go
awgreene: I support this change, but recommend doing so in a separate commit.

conditions/conditions.go
awgreene: I support this change, but recommend doing so in a separate commit.

awgreene: See https://github.com/operator-framework/operator-lib/pull/65

conditions/conditions.go
awgreene: See https://github.com/operator-framework/operator-lib/pull/65

varshaprasad96 (approved):
/lgtm
Thanks @awgreene !

PR #65, opened 2021-07-06
Depreciate ErrNoOperatorCondition

awgreene:
Problem: The ErrNoOperatorCondition does not surface any errors
encountered when retrieving OperatorConditions from the cluster.
It is not possible to know if there is an RBAC issue, if the
resource does not exist, or some other error.

Solution: Surface the error retrieving the OperatorCondition

### DIFF
--- a/conditions/conditions.go
+++ b/conditions/conditions.go
@@ -29,6 +29,7 @@ import (
 
 var (
 	// ErrNoOperatorCondition indicates that the operator condition CRD is nil
+	// Depreciated
 	ErrNoOperatorCondition = fmt.Errorf("operator Condition CRD is nil")
 
 	// readNamespace gets the namespacedName of the operator.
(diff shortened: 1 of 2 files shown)


varshaprasad96 (approved):
/lgtm

joelanford:
Just a belated nit: Since we no longer return this error, I think we should do one more thing.
Either:
- Add a bit more the the `Deprecated` comment (and fix the misspelling so it shows up in IDEs) to explain that this error is no longer returned and users should always expect the more generic error from the client
- Just go ahead and remove it (we're not v1 in operator-lib yet, so breaking changes are alright).

awgreene:
> Just go ahead and remove it (we're not v1 in operator-lib yet, so breaking changes are alright).
+1

awgreene:
@joelanford I created https://github.com/operator-framework/operator-lib/pull/66 to remove the ErrNoOperatorCondition.

estroz:
A 3rd option: wrap the underlying error in ErrNoOperatorCondition, which is my preference.

awgreene:
@estroz what value did we get from the `ErrNoOperatorCondition` error?

estroz:
I don't know, I was assuming there was a particular reason it was added in the first place. If you don't think there is benefit, then I'm in favor of removing the error entirely.

joelanford:
It seems like the intention was to signal `not found`, but I agree with @awgreene's premise that:
a) it was swallowing and translating non-`non found` errors, which could give callers an incorrect impression
b) callers can just use `apierrors.IsNotFound()` to test the equivalent of what `ErrNoOperatorCondition` seems like it was meant to convey.

PR #68, opened 2021-07-27
[WIP] feat(pattern): add job harvester pattern

estroz:
**Description of the change:** added the `pattern/jobharvester` package with Job log forwarding and garbage collection functionality.

**Motivation for the change:** lots of operators spawn jobs for various tasks. This simple controller exposes both the primitive `Harvester` type to get a Job's Pod logs then delete the Pod, or a higher level controller implementation for low-LOC, background Job harvesting.

/kind feature

Signed-off-by: Eric Stroczynski <ericstroczynski@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/operator-framework/api v0.10.0
 	github.com/prometheus/client_golang v1.11.0
 	github.com/prometheus/client_model v0.2.0
+	go.uber.org/zap v1.17.0
 	k8s.io/api v0.21.1
 	k8s.io/apimachinery v0.21.1
 	k8s.io/client-go v0.21.1
--- a/pattern/jobharvest/clientgo.go
+++ b/pattern/jobharvest/clientgo.go
@@ -1,0 +1,21 @@
+package jobharvest
+
+import (
+	"k8s.io/client-go/kubernetes"
+)
+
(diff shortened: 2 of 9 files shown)


estroz:
@tlwu2013 @jmccormick2001 whipped up this super simple (functionality-wise) POC after our discussion yesterday. All this does is write a Job's Pod logs streams to some writer, then delete the Job. I'd like to try out some ideas for log writer interfaces as the current `LogReader` API is rudimentary. Let me know if you have any!

estroz:
There is an alternative to this implementation, which does not involve adding a new controller/helper for cleaning up jobs, that imo is a more desirable approach but involves more input from library users:
1. Provide a helper function that appends a log collection container to a Job's Pod template that can access all container logs via either the downward API or some other mechanism. This container must have authorization to connect to and be able to authenticate with a log server that receives logs.
1. Use the [`ttlSecondsAfterFinished`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs) field to garbage collect the Job and its child resources once logs are collected.

Implementing (1) is the difficult bit, since a custom log-forwarding image is required.

estroz:
After some thought, it seems the controller approach is the right way to go for a general solution. The TTL controller approach works great if you have logging infrastructure that you can drop into a pod spec, which many operators do; the logging implementation is difficult to create a pattern around since usage varies widely depending on what log aggregation system is used.

PR #70, opened 2021-07-28
Fix permission issue when settin the operator-condition

### LINKED ISSUE #69: Permission error when trying to set the operator condition
## Bug Report

**What did you do?**
Trying to use operator-condition in HCO. Getting errors in the log when trying to set the condition:

> operatorconditions.operators.coreos.com "kubevirt-hyperconverged-operator.v1.5.0" is forbidden: User "system:serviceaccount:kubevirt-hyperconverged:hyperconverged-cluster-operator" cannot update resource "operatorconditions/status" in API group \"operators.coreos.com\" in the namespace "kubevirt-hyperconverged"

The operator-condition is set to the required value

**Bug source**
The library still uses `c.client.Status().Update()` instead of `c.client.Update()`. 

https://github.com/operator-framework/operator-lib/blob/main/conditions/conditions.go#L101

**Environment**
* operator-lib version: `0.50.0`

**Possible Solution**

**Additional context**

nunnatsa:
The conditions moved from the `Status` field to the `Spec`, but when setting the condition, the code still uses `client.Status().Update()` to set the condition, instead of just `client.Update()`. This is wrong in two ways: first, the code is trying to update the spec, so it will fail, and also, the operator does not have permissions to update the status. As result, the `Set` function always returns error.

This PR fixes #69.

Signed-off-by: Nahshon Unna-Tsameret <nunnatsa@redhat.com>

This PR changes the client call to use the regular client instead of the status client.

Can't use the current version of the library to set an operator condition. This PR will allow using the lib.

### DIFF
--- a/conditions/conditions.go
+++ b/conditions/conditions.go
@@ -98,7 +98,7 @@ func (c *condition) Set(ctx context.Context, status metav1.ConditionStatus, opti
 		}
 	}
 	meta.SetStatusCondition(&operatorCond.Spec.Conditions, *newCond)
-	err = c.client.Status().Update(ctx, operatorCond)
+	err = c.client.Update(ctx, operatorCond)
 	if err != nil {
 		return err
 	}


awgreene:
@nunnatsa thanks for catching this - could you fix the commit message header? "Permission" is misspelt.

awgreene:
Also - 
This PR **changes** the client call to use the regular client instead of the status client.

nunnatsa:
@awgreene - done

varshaprasad96 (approved):

PR #71, opened 2021-08-12
Add function to facilitate proxy-aware operators

asmacdo:
Signed-off-by: austin <austin@redhat.com>

Adds a helper function to retrieve proxy-related environment variables.

As a best practice, operators should be configurable (via the environment variables set on the operator itself) to set up operands that use a network proxy.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -24,7 +24,7 @@ import (
 // ProxyEnvNames are standard environment variables for proxies
 var ProxyEnvNames = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 
-// ReadProxyVarssFromEnv retrieves the standard proxy-related environment
+// ReadProxyVarsFromEnv retrieves the standard proxy-related environment
 // variables from the running environment and returns a slice of corev1 EnvVar
 // containing upper and lower case versions of those variables.
 func ReadProxyVarsFromEnv() []corev1.EnvVar {
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -30,16 +30,18 @@ var ProxyEnvNames = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 func ReadProxyVarsFromEnv() []corev1.EnvVar {
 	envVars := []corev1.EnvVar{}
 	for _, s := range ProxyEnvNames {
-		value := os.Getenv(s)
-		envVars = append(envVars, corev1.EnvVar{
-			Name:  s,
-			Value: value,
-		})
-		// Duplicate values for upper and lower case
-		envVars = append(envVars, corev1.EnvVar{
-			Name:  strings.ToLower(s),
-			Value: value,
-		})
+		value, isSet := os.LookupEnv(s)
+		if isSet {
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  s,
+				Value: value,
+			})
+			// Duplicate values for upper and lower case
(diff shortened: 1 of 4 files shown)


asmacdo:
From https://github.com/operator-framework/enhancements/pull/89

proxy/proxy_test.go
```diff
@@ -0,0 +1,65 @@
...
+			return envVars[i].Value, nil
+		}
+	}
+	return "", errors.New("empty name")
+
+}
+
+var _ = Describe("Retrieving", func() {
+	Describe("proxy environment variables", func() {
+		// TODO(asmacdo) seems like a bad idea to mess with the env
```
asmacdo, line 39: Hold placed to remove this after I get some feedback, the unit test is currently weak.

proxy/proxy.go
```diff
@@ -0,0 +1,45 @@
...
+// ProxyEnvNames are standard environment variables for proxies
+var ProxyEnvNames = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+
+// ReadProxyVarsFromEnv retrieves the standard proxy-related environment
+// variables from the running environment and returns a slice of corev1 EnvVar
+// containing upper and lower case versions of those variables.
+func ReadProxyVarsFromEnv() []corev1.EnvVar {
+	envVars := []corev1.EnvVar{}
+	for _, s := range ProxyEnvNames {
+		value := os.Getenv(s)
```
varshaprasad96, line 33: does this mean that its upto the user of this library to verify if the variable is empty or not? In the sense, even if the variables are not set, the value is appended in `[]corev1.EnvVar`. Do we need to add a check to omit the ones which are not set, or was this intentional ?

asmacdo: Good catch, I was just setting it to empty string in that case, but I can use `LookupEnv` to only pass through only when the var is set.

proxy/proxy.go
```diff
@@ -0,0 +1,47 @@
...
+		if isSet {
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  s,
+				Value: value,
+			})
+			// Duplicate values for upper and lower case
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  strings.ToLower(s),
+				Value: value,
+			})
```
varshaprasad96, line 43:
```suggestion
			envVars = append(envVars, corev1.EnvVar{
				Name:  s,
				Value: value,
			}, corev1.EnvVar{
				Name:  strings.ToLower(s),
				Value: value,
			})
```

varshaprasad96: is there specific reason to have lower and upper case appended? Can we always convert values to one case and add it to slice? Would it be easier to have a getter that checks and returns the specific value for the variable asked. With this we need not store two values, and user can obtain the the specific value for the key they specify. 

```go
func GetProxyValue(input string) (string, error) {
    // always convert to lower case so that we don't have to store two keys for same value
     key := strings.ToLower(input)
     if val, ok := envVars[key]; ok {
             return val, nil
     }
     return "",  fmt.Errorf("%s not set", input)
}
```
With this we may or may not want to expose `envVars` (the slice of proxy variables entirely). Its upto us. And if we want we can go a step further by making `ProxyEnvNames` to be a particular type.

If there is a specific reason to have these keys in both the cases and to return the list entirely, please feel free to ignore the comment.

varshaprasad96 (approved):

jmrodri (approved):
/lgtm
/approve

PR #77, opened 2021-11-10
Remove locks for Pods in "Shutdown" status

### LINKED ISSUE #73: Leader election cannot complete after node reboot on vSphere
## Bug Report
Leader election function fails to detect non-running leader pod on Ready node.

**What did you do?**
Rebooted several nodes of vSphere 7 Kubernetes cluster. After some time, all nodes went up and Ready. Operator pods, which existed before reboot, stuck with ProviderFailed status reason and Ready condition = false.

Operator pods restarted by deployment, or removed by leader election function. Leader election successful for one of pods.

**What did you see instead? Under which circumstances?**
New pods created and all waiting for leader lock go off. Previous leader pod is not evicted, and as its node is Ready, leader deletion never happens.

**Environment**
* operator-lib version: v0.6.0
* vSphere 7 U3

**Possible Solution**
If current leader pod is not Ready after some significant timeout (2-3 minutes, maybe configurable), perform delete and re-elect.

**Additional context**
ProviderFailed status is platform-specific (and is likely to be a bug), but operator is expected to detect failure based on pod conditions.

scheying:
Kubernetes introduced ["Graceful Node Shutdown"](https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown) as an Alpha feature in 1.20. Pods that are evicted by this mechanism have their status.reason attribute set to "Shutdown". This commit extends the garbage collection so that locks for Pods in the "Shutdown" state are also released.

Deadlocks caused by Pods in "shutdown" state still holding leader locks were observed on a live GKE Cluster with preemtible nodes.

This might also fix #73.

### DIFF
--- a/leader/leader.go
+++ b/leader/leader.go
@@ -229,7 +229,7 @@ func myOwnerRef(ctx context.Context, client crclient.Client, ns string) (*metav1
 
 func isPodEvicted(pod corev1.Pod) bool {
 	podFailed := pod.Status.Phase == corev1.PodFailed
-	podEvicted := pod.Status.Reason == "Evicted"
+	podEvicted := pod.Status.Reason == "Evicted" || pod.Status.Reason == "Shutdown"
 	return podFailed && podEvicted
 }
 


witsch:
@fabianvf @joelanford would you mind having a look, please? The change looks pretty straightforward (and very small anyway) and the issue also exists for us… 😊

PR #100, opened 2022-01-25
Closes #99: Make logging configurable.

### LINKED ISSUE #99: Prune logging cannot be set
## Bug Report

**What did you do?**

Wanting to [configure the Prune structure](https://sdk.operatorframework.io/docs/best-practices/resource-pruning/#operator-lib-prune-library) but the [log field](https://github.com/operator-framework/operator-lib/blob/main/prune/prune.go#L73) is not exported and never set.

Either that the field is automatically populated or that it is exported.

**What did you see instead? Under which circumstances?**

A field that is not exported and not automatically set.

**Possible Solution**

Although, I am not in favor of using global variables for loggers things should be consistent inside a library. The other features provided by operator-lib for instance [event handler](https://github.com/operator-framework/operator-lib/blob/main/handler/enqueue_annotation.go#L32) and [leader election](https://github.com/operator-framework/operator-lib/blob/main/leader/leader.go#L43) use a preset variable:
~~~
var log = logf.Log.WithName("event_handler")
var log = logf.Log.WithName("leader")
~~~

logf is a package of the controller-runtime library, which contains logr.Logger Log. This assumes that SetLogger was called earlier, which is the same expectation sets by Kubebuilder/controller-runtime. 
This makes this library less suitable for clients or other apps not based on controller-runtime but again things should be consistent.

Another thing I would like to mention and a second issue could get open for it is that the naming/structure could be improved. It seems a bit unconventional to have a method like Execute on a structure named Config. I would expect that more on a structure called Pruner. Config being what is provided by users to get Pruners that fits their needs.

fgiloux:

Closes #99
This pull request aligns the logging configuration in the prune feature with what is done in the rest of the library. It declares a package level variable for a logger. The logger is retrieved from the controller-runtime library, which is leveraging logr.

The current prune implementation could not be used outside of the prune package as the log field of the Config struct was unexported.

A pull request for amending the documentation will follow.

### DIFF
--- a/prune/prune.go
+++ b/prune/prune.go
@@ -23,6 +23,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
+
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // ResourceStatus describes the Kubernetes resource status we are evaluating
@@ -54,10 +56,10 @@ type StrategyConfig struct {
 
 // StrategyFunc function allows a means to specify
 // custom prune strategies
-type StrategyFunc func(cfg Config, resources []ResourceInfo) ([]ResourceInfo, error)
+type StrategyFunc func(ctx context.Context, cfg Config, resources []ResourceInfo) ([]ResourceInfo, error)
 
 // PreDelete function is called before a resource is pruned
-type PreDelete func(cfg Config, something ResourceInfo) error
+type PreDelete func(ctx context.Context, cfg Config, something ResourceInfo) error
 
 // Config defines a pruning configuration and ultimately
 // determines what will get pruned
@@ -70,13 +72,13 @@ type Config struct {
 	Strategy       StrategyConfig            //strategy for pruning, either age or max
 	CustomStrategy StrategyFunc              //custom strategy
 	PreDeleteHook  PreDelete                 //called before resource is deleteds
-	log            logr.Logger
+	Log            logr.Logger               //optional: to overwrite the logger set at context level
 }
 
 // Execute causes the pruning work to be executed based on its configuration
 func (config Config) Execute(ctx context.Context) error {
-
-	config.log.V(1).Info("Execute Prune")
+	log := Logger(ctx, config)
+	log.V(1).Info("Execute Prune")
 
 	err := config.validate()
 	if err != nil {
@@ -92,13 +94,13 @@ func (config Config) Execute(ctx context.Context) error {
 			if err != nil {
 				return err
 			}
-			config.log.V(1).Info("pods ", "count", len(resourceList))
+			log.V(1).Info("pods ", "count", len(resourceList))
 		} else if config.Resources[i].Kind == JobKind {
 			resourceList, err = config.getCompletedJobs(ctx)
 			if err != nil {
 				return err
 			}
-			config.log.V(1).Info("jobs ", "count", len(resourceList))
+			log.V(1).Info("jobs ", "count", len(resourceList))
 		}
 
 		var resourcesToRemove []ResourceInfo
@@ -109,7 +111,7 @@ func (config Config) Execute(ctx context.Context) error {
 		case MaxCountStrategy:
 			resourcesToRemove, err = pruneByMaxCount(ctx, config, resourceList)
 		case CustomStrategy:
-			resourcesToRemove, err = config.CustomStrategy(config, resourceList)
+			resourcesToRemove, err = config.CustomStrategy(ctx, config, resourceList)
 		default:
 			return fmt.Errorf("unknown strategy")
 		}
@@ -123,7 +125,7 @@ func (config Config) Execute(ctx context.Context) error {
 		}
 	}
 
-	config.log.V(1).Info("Prune completed")
+	log.V(1).Info("Prune completed")
 
 	return nil
 }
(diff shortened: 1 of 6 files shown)


fgiloux:
cc @camilamacedo86 @Sijoma @ryantking

fgiloux:
@camilamacedo86 can you please review the PR?

camilamacedo86 (commented):
/lgtm

camilamacedo86 (approved):

fgiloux:
@theishshah @varshaprasad96 Anything preventing the merge of this PR?

joelanford:
@fgiloux I made [this comment](https://github.com/operator-framework/operator-lib/issues/99#issuecomment-1074145194) over on the issue. WDYT?

fgiloux:
@joelanford @varshaprasad96 @camilamacedo86 reworked to address Joe's concerns

prune/prune.go
```diff
@@ -181,3 +181,17 @@ func (config Config) validate() (err error) {
 	}
 	return nil
 }
+
+// Logger returns a logger from the context using logr method or Config.Log if none is found
+// controller-runtime automatically provides a logger in context.Context.
+// Note that there is no compile time check whether a logger can be retrieved by either way.
+// keysAndValues allow to add fields to the logs, cf logr documentation.
+func Logger(ctx context.Context, cfg Config, keysAndValues ...interface{}) logr.Logger {
+	log := cfg.Log
```
joelanford, line 190: I kind of feel like `cfg.Log` (if it's set) should win over `logr.FromContext()`. WDYT?

prune/prune.go
```diff
@@ -181,3 +181,17 @@ func (config Config) validate() (err error) {
...
+	if ctx != nil {
+		if logger, err := logr.FromContext(ctx); err == nil {
```
joelanford, line 192: Should we use controller-runtime's version of this function so that if there isn't a logger in the context, we end up with the controller-runtime's global logger? Or is that too much coupling to controller-runtime?

prune/prune.go
```diff
@@ -181,3 +181,17 @@ func (config Config) validate() (err error) {
 	}
 	return nil
 }
+
+// Logger returns a logger from the context using logr method or Config.Log if none is found
+// controller-runtime automatically provides a logger in context.Context.
```
joelanford, line 186:
```suggestion
// controller-runtime automatically provides a logger in context.Context during Reconcile calls.
```
Maybe other times as well? This is the only one I know about though.

prune/prune.go
```diff
@@ -181,3 +181,17 @@ func (config Config) validate() (err error) {
...
+// Note that there is no compile time check whether a logger can be retrieved by either way.
+// keysAndValues allow to add fields to the logs, cf logr documentation.
+func Logger(ctx context.Context, cfg Config, keysAndValues ...interface{}) logr.Logger {
```
joelanford, line 189: Does this function need to be exported?

prune/prune.go
fgiloux: My thinking was that cfg.Log is set once whereas logr.FromContext() can be changed, although unlikely with every request. I picked the one with the scope having the shorter lifetime

prune/prune.go
fgiloux: My thinking was to avoid coupling with controller-runtime... but the library has already a dependency on controller-runtime. I think Jeff wanted to make it available to other frameworks, hence 
~~~
type Config struct {
        Clientset      kubernetes.Interface      // kube client used by pruning
~~~
Due to the already existing dependency I can use the controller-runtime version

prune/prune.go
fgiloux: I won't debate on that :-)

prune/prune.go
fgiloux: It is exported in controller-runtime and the user can leverage it for custom code that gets injected through PreDeleteHook and CustomStrategy. I am passing the context to them for the purpose.

prune/prune.go
fgiloux: This also means that we will always get a logger from context. In that case this cannot have precedence over the logger set in the config struct otherwise this would never be set.

prune/prune.go
joelanford: My thought was that if I go to the trouble of setting the logger in the config, that's an extra explicit step I'm taking to say "use this logger". With the code as is, if I was using this in `Reconcile`, I'd have to create a new context that drops the logger from the Reconcile context before using the prune functions. That seems pretty awkward.

prune/prune.go
joelanford: Got it. Makes sense. Thanks for the explanation!

camilamacedo86:
Hi @joelanford @jmrodri @varshaprasad96 

This project still < 1.0.0, any reason for we do not move forward when the api diff fails?
Do we need to ensure backwords compatibility in the lib?

joelanford:
Do we know what the impact would be if this broke (what change is necessary for callers?, how many operators are using the library?). I'd say its okay to break this if necessary, but we should generally try to avoid it. 

Did someone mention that @ryantking had a larger design or some follow-on changes to make? If so, I'd suggest that we try to avoid breaking these APIs twice.

fgiloux:
@camilamacedo86 @joelanford as per #99 the current situation is that Config.log is not exported and not set automatically but it is [called in the code](https://github.com/operator-framework/operator-lib/blob/main/prune/maxcount.go#L35). In my opinion this makes the library broken. Hence there is nothing to break. As the prune library is documented in multiple places the urgency is to make it usable That's the aim of this PR.
To answer Joe''s questions:

> Do we know what the impact would be if this broke (what change is necessary for callers?

No impact, adding a field, which is optional is not breaking an API. This means if the library had happened to work it would work further but it did not by any way.

> how many operators are using the library?). I'd say its okay to break this if necessary, but we should generally try to avoid it.

As log is called and not configurable my guess would be exactly 0.

> Did someone mention that @ryantking had a larger design or some follow-on changes to make? If so, I'd suggest that we try to avoid breaking these APIs twice.

I mentioned that. The [EP](https://github.com/operator-framework/enhancements/pull/109) has been however stale waiting to get merged for months. The urgency is to make a documented library usable.

joelanford:
Seems like the breaking change is adding a context parameter to the PreDelete and StrategyFunc function signatures.

Based on everything @fgiloux said, it seems like a no brainer to get this library to a working state even if that means some minor breaking API changes.

/lgtm

PR #104, opened 2022-02-27
mod: Upgrade to ginkgo v2

haoxins:


camilamacedo86:
It will not be compatible with the version used in SDK/KB and controller runtime.
So, could it not bring any problem for the users?
Also, it has breaking changes. Are we ok with introduces breaking changes here?

PR #105, opened 2022-04-08
Feature/auto pruning

### LINKED ISSUE #101: Improve prune feature implementation
## Feature Request

**Is your feature request related to a problem? Please describe.**
It seems that  the implementation of the prune feature could be improved. As part of the fix for PR #100 I noticed the following:
- Naming issue: Config is the main struct for the prune logic, which leads to methods like:
`Config.Execute`. It does not sound right. Config may be passed to a factory creating a Pruner or a Processor structure but the Execuste method should be on the later structure: Pruner.Execute() or Processor.Execute() makes more sense.
- The Config structure seems not optimal in the sense that  for instance the same Strategy has to be applied to all the resources. If the logic is to be configured once for multiple resources it may make more sense to have a slice or a map of substructures containing the specificity for each resource: gvk, namespace, strategy, pre-deletion hook. A different strategy can then be selected by the user for each resource.
- The ResourceInfo structure seems to be limiting. Using [Unstructured](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured) would allow working with any type. The additional logic would then be able to work with fields specific to the type being processed.
- The pod deletion strategy only considers succeeded pods. It should also consider failed pods. Basically all pods with the [terminated state](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states), possibly letting the user selects the phases that are candidates for pruning.
- It may make more sense to use a closure for strategy rather than a structure strategyConfig and a func StrategyFunc. By doing so it would be possible to unify the interface for what is implemented by the library: pruneByMaxAge, pruneByMaxCount and what is provided by the user: CustomStrategy. They would all be a func, possibly passed as a Config field with this signature:
  ~~~
  func pruneStrategyt(resources []unstructured.Unstructured) (resourcesToRemove [][]unstructured.Unstructured, err error)
  ~~~
  Specific parameters part of the Config would be made available through closure variables. pruneByMaxCount would then not need a   Config.strategy, which provides information meant for all possible strategies. It would make things easier to extend and the implementation would be more performant: No need for switch config.Strategy.Mode in prune.go. All the prune strategies would have the same signature.
- The context is passed to the functions in the library but my feeling is that we have two use cases:
  - The prune logic is called from a reconciliation loop and the context may provide more information
  - The prune logic is called from a cron job (as per the documentation) and the context will provide little information
 I haven't seen it really used. Is it thought more as a safety net (if anything is needed that was not foreseen get it through the context)? Can this be addressed through the closure approach?
- We should also aim and I believe the points listed above would help with that not to limit the logic to pods and jobs.

everettraven:
Update the pre-existing pruning functionality

- Issue #101
- @ryantking EP: https://github.com/operator-framework/enhancements/pull/109

resolves #101


prune/prunables.go
```diff
@@ -0,0 +1,104 @@
...
+	// if the object is not a Pod then it is not pruneable
+	if obj.GetObjectKind().GroupVersionKind() != gvk {
+		return fmt.Errorf("can not prune object as it is not a Pod")
+	}
+
+	// convert to a Pod object so we can do Pod specific checks
+	pod := &corev1.Pod{}
+	err := ConvertClientObjectToKind(obj, pod)
+
+	if err != nil {
```
jmrodri, line 48: nit: drop the space between 46 and 48.

prune/prunables.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if obj.GetObjectKind().GroupVersionKind() != gvk {
+		return fmt.Errorf("can not prune object as it is not a Job")
+	}
+
+	job := &batchv1.Job{}
+	err := ConvertClientObjectToKind(obj, job)
+
+	if err != nil {
```
jmrodri, line 77: nit: drop the space between 75 and 77

jmrodri: Same is true anywhere there is an `if err ~= nil` handler. Seems more logical that it is with the line immediately preceding it.

prune/prune.go
```diff
@@ -17,181 +17,197 @@ package prune
...
 }
 
-// Execute causes the pruning work to be executed based on its configuration
-func (config Config) Execute(ctx context.Context) error {
-	log := Logger(ctx, config)
-	log.V(1).Info("Execute Prune")
+// StrategyFunc takes a list of resources and returns the subset to prune.
+type StrategyFunc func(ctx context.Context, objs []client.Object) ([]client.Object, error)
+
+// IsPruneableFunc is a function that checks a the data of an object to see whether or not it is safe to prune it.
```
jmrodri, line 74:
```suggestion
// IsPruneableFunc is a function that checks the data of an object to see whether or not it is safe to prune it.
```

prune/prune.go
```diff
@@ -17,181 +17,197 @@ package prune
...
+// It should return `nil` if it is safe to prune, `ErrUnpruneable` if it is unsafe, or another error.
```
jmrodri, line 75: I don't see anything returning `ErrUnpruneable`. Either the methods should return this error or we should update the comment to reflect what is really being returned.

prune/prune.go
```diff
@@ -17,181 +17,197 @@ package prune
...
+// It should safely assert the object is the expected type, otherwise it might panic.
+type IsPruneableFunc func(obj client.Object) error
+
+// PrunerOption configures the pruner.
+type PrunerOption func(p *Pruner)
+
+// Error returns a string reprenstation of an `ErrUnpruneable` error.
```
jmrodri, line 82:
```suggestion
// Error returns a string representation of an `ErrUnpruneable` error.
```

jmrodri (changes requested):
first round. A few nits, a couple of typos, and a question about ErrUnprunable.

prune/prune.go
everettraven: fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/1fca33a0903e9f38f9f39390c315c8618bbc2414

prune/prune.go
everettraven: fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/1fca33a0903e9f38f9f39390c315c8618bbc2414

prune/prunables.go
everettraven: Should all be fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/1fca33a0903e9f38f9f39390c315c8618bbc2414

prune/prunables.go
everettraven: fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/1fca33a0903e9f38f9f39390c315c8618bbc2414

prune/prunables.go
```diff
@@ -0,0 +1,114 @@
...
+		Kind:    "Pod",
+	}
+
+	// if the object is not a Pod then it is not pruneable
+	if obj.GetObjectKind().GroupVersionKind() != gvk {
+		return &ErrUnprunable{
+			Obj:    &obj,
+			Reason: "Object is not a Pod",
+		}
+	}
```
ryantking, line 45: You're checking the Kubernetes type, but, while related, we care about the go type here. So you'd want something like:

```go
pod, ok := obj.(*corev1.Pod)
if !ok {
    return // a cool error
}
```

To add more detail:

```go 
var pod corev1.Pod
gvk := pod.GetObjectKind().GroupVersionKind()
```

After running that snippet, `gvk` will be an empty `GroupVersionKind` struct since its not set in the zero value Pod struct. This probably wouldn't happen in real life since you'll likely be reading the Pod from the server, which will return it with correct GVK information, but it demonstrates when the Kubernetes type and actual go type can drift.

One more thing, this is one of the rare situations that I think a `panic` or `log.Fatal` would be a better design than returning an error. Usually `panic`-ing in go is a bad practice, but I think this fits the narrow slice of usecases where it makes sense. Typically a `panic` means "something has happened that the program cannot recover from." If an operator-author incorrectly registers a pruneable function for the wrong GVK, that would be something that the program cannot recover from since it cannot correct a mistake the user made when configuring it. It also wouldn't make sense in my opinion to expose an error since someone consuming the operator is unable to fix that error since the code itself is broken and needs to be fixed. I would keep it simple and just use this:

```go
pod := obj.(*corev1.Pod)
```

A type-assertion will panic if it is not possible and there is only one return value. I think in this case, we are fine letting the program panic if the pod isPruneable func gets a non-pod object.

prune/prunables.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// convert to a Pod object so we can do Pod specific checks
+	pod := &corev1.Pod{}
+	err := ConvertClientObjectToKind(obj, pod)
+	if err != nil {
+		return err
+	}
```
ryantking, line 52: This will be handled by the type assertion.

prune/prunables.go
```diff
@@ -0,0 +1,114 @@
...
+			Obj:    &obj,
+			Reason: "Object is not a Job",
+		}
+	}
+
+	job := &batchv1.Job{}
+	err := ConvertClientObjectToKind(obj, job)
+	if err != nil {
+		return err
+	}
```
ryantking, line 86: Same comment from above

prune/prunables.go
```diff
@@ -0,0 +1,114 @@
...
+		Version: "v1",
+		Kind:    "Job",
+	}
+
+	if obj.GetObjectKind().GroupVersionKind() != gvk {
+		return &ErrUnprunable{
+			Obj:    &obj,
+			Reason: "Object is not a Job",
+		}
+	}
```
ryantking, line 80: Something else to call out about the usage of `ErrUnprunable`. It is not intended to wrap actual errors. It means "this resource should not be pruned due to the business logic of the program."

prune/prune.go
```diff
@@ -16,184 +16,204 @@ package prune
...
-	LabelSelector  string                    //selector resources to prune
-	DryRun         bool                      //true only performs a check, not removals
-	Resources      []schema.GroupVersionKind //pods, jobs are supported
-	Namespaces     []string                  //empty means all namespaces
-	Strategy       StrategyConfig            //strategy for pruning, either age or max
-	CustomStrategy StrategyFunc              //custom strategy
-	PreDeleteHook  PreDelete                 //called before resource is deleteds
-	Log            logr.Logger               //optional: to overwrite the logger set at context level
+// ErrUnprunable indicates that it is not allowed to prune a specific object.
+type ErrUnprunable struct {
```
ryantking, line 66: I wonder if we should drop the `Err` prefix here and just do `Unrpuneable`. Go uses this pattern internally with things like `io.EOF` that aren't necessarily errors, but indicates that an operation cannot continue.

prune/prune.go
```diff
@@ -16,184 +16,204 @@ package prune
...
+	Reason string
 }
 
-// Execute causes the pruning work to be executed based on its configuration
-func (config Config) Execute(ctx context.Context) error {
-	log := Logger(ctx, config)
-	log.V(1).Info("Execute Prune")
+// Error returns a string representation of an `ErrUnprunable` error.
+func (e *ErrUnprunable) Error() string {
+	return fmt.Sprintf("%s -- %+v", e.Reason, e.Obj)
```
ryantking, line 73: In general, I like to keep `Error` returns to one line since they will likely be logged, and multiline things make log messages hard to read. This is specifically in regards to the "%+v" format directive for the object, which will dump the whole struct in a way that is pretty hard to use since go's default formatting for complex types is not easy on the eyes.

What I would like is the method here to return a human-readable one-liner:

```go
    return fmt.Sprintf("unable to prune %s: %s", client.ObjectKeyFromObject(e.Obj), e.Reason)
```

`client.ObjectKey` has a `String() string` method  that returns `namespace/name`. The `%s` format directive will automatically call it if present.

Then, where the error actually occurs, you can add a debug log call that dumps the entire object. That way by logs will only have the human-readable description and users can turn on debug logging if they want look at the values of the object. This is a common pattern:

```go
if err != nil {
    log.Debug("a dump of a large struct")
    return err
}
```

prune/prune.go
```diff
@@ -16,184 +16,204 @@ package prune
...
+}
+
+// StrategyFunc takes a list of resources and returns the subset to prune.
+type StrategyFunc func(ctx context.Context, objs []client.Object) ([]client.Object, error)
```
ryantking, line 77: Can you see if `[]client.ObjectKey` is sufficient for the first return type?

prune/prune.go
```diff
@@ -16,184 +16,204 @@ package prune
...
-				return err
-			}
-			log.V(1).Info("jobs ", "count", len(resourceList))
-		}
+	// Populate the default IsPrunableFunc(s)
+	RegisterIsPrunableFunc(schema.GroupVersionKind{
+		Group:   "core",
+		Version: "v1",
+		Kind:    "Pod",
+	}, DefaultPodIsPrunable)
```
ryantking, line 106: These GVKs should be defined somewhere in the kubernetes API. I can help you look if you can't find them, but I don't remember where exactly they are off the top of my head.

prune/registry.go
```diff
@@ -0,0 +1,46 @@
...
+	// Pruneables is a map of GVK to an IsPrunableFunc
+	Pruneables map[schema.GroupVersionKind]IsPrunableFunc
+}
+
+func NewRegistry() *Registry {
+	return new(Registry)
+}
+
+var DefaultRegistry = &defaultRegistry
+
```
ryantking, line 32: This should be a function, exported globals have a very narrow usecase.

prune/registry.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	r.Pruneables[gvk] = IsPrunable
+}
+
+// RegisterIsPrunableFunc registers a function to check whether it is safe to prune a resource of a certain type.
+func RegisterIsPrunableFunc(gvk schema.GroupVersionKind, IsPrunable IsPrunableFunc) {
+	if DefaultRegistry.Pruneables == nil {
+		DefaultRegistry.Pruneables = make(map[schema.GroupVersionKind]IsPrunableFunc)
+	}
+	DefaultRegistry.Pruneables[gvk] = IsPrunable
```
ryantking, line 48: This should just call:

```go
    defaultRegistry.RegisterIsPruneableFunc(gvk, isPruneable)
```

prune/registry.go
```diff
@@ -0,0 +1,46 @@
...
+func NewRegistry() *Registry {
+	return new(Registry)
+}
+
+var DefaultRegistry = &defaultRegistry
+
+var defaultRegistry Registry
+
+// RegisterIsPrunableFunc registers a function to check whether it is safe to prune a resource of a certain type.
+func (r *Registry) RegisterIsPrunableFunc(gvk schema.GroupVersionKind, IsPrunable IsPrunableFunc) {
```
ryantking, line 36: `IsPruneable` => `isPruneable`

ryantking (changes requested):
Good start, I like the file organization. Would love to see some tests also, even if they're not passing yet.

prune/prunables.go
everettraven: Thanks for the detailed explanation! This makes more logical sense to me after reading your explanation so I will go ahead and work on implementing this for both the `Pod` and `Job` default prunable functions.

prune/prune.go
everettraven: I like this idea, especially because of your previous note that it isn't meant to wrap actual errors. I think removing the `Err` prefix would prevent any confusion regarding whether or not this should be treated as an error. Will work on implementing this.

prune/prune.go
everettraven: I agree with the human-readable one-liner idea, so I will work on implementing this.

prune/prune.go
everettraven: It looks like we should be able to use `[]client.ObjectKey` as the return type, but will have to create an object with a namespace, name, and GVK set like so (code taken from https://github.com/kubernetes-sigs/controller-runtime/blob/0a4d154ec52c0359f4b20ec9e2845a2e7af125a0/pkg/client/example_test.go#L221-L230):
```go
	// Using a unstructured object.
	u := &unstructured.Unstructured{}
	u.SetName("name")
	u.SetNamespace("namespace")
	u.SetGroupVersionKind(schema.GroupVersionKind{
		Group:   "apps",
		Kind:    "Deployment",
		Version: "v1",
	})
	_ = c.Delete(context.Background(), u)
```

I am going to take a shot at it and see if it will work how we are expecting.

prune/prune.go
everettraven: Replaced this (and all other manual GVK creation of GVKs that should already exist) with:
```go
podGVK := corev1.SchemeGroupVersion.WithKind("Pod")
```
Let me know what you think

everettraven:
Mentioning @fgiloux due to involvement in the review of the EP for this feature.

ryantking (changes requested):
I think we need to discuss the `Pruner` struct. Using functional arguments AND exposing fields kind of strikes me as an anti-pattern because you could have something like this:

```go
func main() {
    pruner := NewPruner(client.New())
    doEvilThings(p)
    p.Prune(ctx) // this will blow up calling a nil client
}

func doEvilThings(p *Pruner) {
    p.Client = nil
}
```

When a struct has publicly modifiable fields, the methods cannot trust the fields to be valid so much validate them before using them. Look at [this line][https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/request.go;l=573] in `net/http`. Since `request.URL` is publicly modifiable, it must `nil`-check it. Typically, the point of the constructor is to force an instance to be in a good state when its created by having two things:

1. Sane defaults
2. Validation for user overrides.

That way we know that the fields are all either defaults the library chose or overrides the user chose that are valid. I think we should un-export most if not all fields on the `Pruner` and funnel all configuration into the pruner. A rule of thumb I use is if the point of the struct is to hold structured data like a Kubernetes type then export the fields, but if the point of the struct is to do work, then it should have a constructor that enforces a good state.

Generally speaking, I'm happy with how this is shaping up. I will review the tests once we get the main library to a good state. Great work.

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
+type Pruner struct {
+	Registry
+
+	// Client is the k8s client that will be used
+	Client client.Client
+
+	// DryRun indicates whether or not we should actually perform pruning or just return the list of pruneable objects
+	// true = Just check, don't prune
+	// false (default) = Prune
+	DryRun bool
```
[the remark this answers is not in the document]
everettraven: I didn't notice that there was a `DryRunClient`, and since we are already having them pass in a client it would probably be better than setting a `DryRun` value on the `Pruner` itself. I think as an added bonus it makes the internal pruner logic more straightforward as we don't have to worry about that `DryRun` parameter. I will go ahead and make the corresponding changes.

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
 
-		err = config.removeResources(ctx, resourcesToRemove)
-		if err != nil {
-			return err
-		}
+// PrunerOption configures the pruner.
+type PrunerOption func(p *Pruner)
+
+// SetRegistry can be used to set the Registry field when configuring a Pruner
+func SetRegistry(registry Registry) PrunerOption {
```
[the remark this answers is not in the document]
everettraven: That makes sense, will go ahead and update all the functional options.

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
+
+	// Prune the resources
+	for _, obj := range objsToPrune {
+		// Prune
+		prunableObj := &unstructured.Unstructured{}
+		prunableObj.SetName(obj.Name)
+		prunableObj.SetNamespace(obj.Namespace)
+		prunableObj.SetGroupVersionKind(p.GVK)
+
+		if err = p.Client.Delete(ctx, prunableObj); err != nil {
```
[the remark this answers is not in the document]
everettraven: This is due to the switch I made from `type StrategyFunc func(ctx context.Context, objs []client.Object) ([]client.Object, error)` --> `type StrategyFunc func(ctx context.Context, objs []client.Object) ([]client.ObjectKey, error)`  in response to a [previous review comment you left](https://github.com/operator-framework/operator-lib/pull/105#discussion_r849668069)

However now that I think about it, I remember us speaking offline about just using the `client.Object` type instead due to us already enforcing use of controller-runtime throughout the `prune` package as well as operator-lib in general.

I think because of this it would be best to switch it back to `type StrategyFunc func(ctx context.Context, objs []client.Object) ([]client.Object, error)` so that it is easier for both the user of the `prune` package and for us to just pass the object in rather than constructing one from a `client.ObjectKey` type. 

What do you think?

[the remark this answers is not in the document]
ryantking: Agreed

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
-		_, err = time.ParseDuration(config.Strategy.MaxAgeSetting)
+	for _, unsObj := range unstructuredObjs.Items {
+		obj, err := convert(p.Client, p.GVK, &unsObj)
 		if err != nil {
-			return err
+			return nil, err
 		}
+
+		if err := p.IsPrunable(obj); isUnprunable(err) {
+			continue
```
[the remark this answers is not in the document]
everettraven: fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/registry.go
```diff
@@ -0,0 +1,67 @@
...
+
+// RegisterIsPrunableFunc registers a function to check whether it is safe to prune a resource of a certain type.
+func RegisterIsPrunableFunc(gvk schema.GroupVersionKind, isPrunable IsPrunableFunc) {
+	DefaultRegistry().RegisterIsPrunableFunc(gvk, isPrunable)
+}
+
+// IsPrunable checks if an object is prunable
+func IsPrunable(obj client.Object) error {
+	return DefaultRegistry().IsPrunable(obj)
+}
```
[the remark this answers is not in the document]
everettraven: fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
+
+	return objsToPrune, nil
 }
 
+/*
+	-------------------------------------------
+	New Auto Pruning API Implementation
+	-------------------------------------------
+*/
+
```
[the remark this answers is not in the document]
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
+	for _, obj := range objsToPrune {
+		// Prune
+		prunableObj := &unstructured.Unstructured{}
+		prunableObj.SetName(obj.Name)
+		prunableObj.SetNamespace(obj.Namespace)
+		prunableObj.SetGroupVersionKind(p.GVK)
+
+		if err = p.Client.Delete(ctx, prunableObj); err != nil {
+			p.Logger.Error(err, "failed to prune resource", "Resource", obj)
+			return nil, fmt.Errorf("failed to prune object -- ERROR -- %s", err)
```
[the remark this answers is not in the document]
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prune.go
```diff
@@ -16,184 +16,249 @@ package prune
...
+	// Prune the resources
+	for _, obj := range objsToPrune {
+		// Prune
+		prunableObj := &unstructured.Unstructured{}
+		prunableObj.SetName(obj.Name)
+		prunableObj.SetNamespace(obj.Namespace)
+		prunableObj.SetGroupVersionKind(p.GVK)
+
+		if err = p.Client.Delete(ctx, prunableObj); err != nil {
+			p.Logger.Error(err, "failed to prune resource", "Resource", obj)
```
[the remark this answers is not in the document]
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prune.go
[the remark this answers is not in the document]
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prune.go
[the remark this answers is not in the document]
everettraven: done with all functional options in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prune.go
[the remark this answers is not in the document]
everettraven: Done in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prunables.go
```diff
@@ -0,0 +1,55 @@
...
+	return nil
+}
+
+// DefaultJobIsPrunable is a default IsPrunableFunc to be used specifically with Job resources.
+// This can be overridden by registering your own IsPrunableFunc via the RegisterIsPrunableFunc method
+func DefaultJobIsPrunable(obj client.Object) error {
+
+	job := obj.(*batchv1.Job)
+
+	// If the job has completed we can remove it
```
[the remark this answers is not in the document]
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prunables.go
```diff
@@ -0,0 +1,55 @@
...
+
+	return nil
+}
+
+// DefaultJobIsPrunable is a default IsPrunableFunc to be used specifically with Job resources.
+// This can be overridden by registering your own IsPrunableFunc via the RegisterIsPrunableFunc method
+func DefaultJobIsPrunable(obj client.Object) error {
+
+	job := obj.(*batchv1.Job)
+
```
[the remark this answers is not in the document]
everettraven: fixed in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

prune/prunables.go
```diff
@@ -0,0 +1,55 @@
...
+package prune
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Some default pruneable functions
```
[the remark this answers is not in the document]
everettraven: removed in https://github.com/operator-framework/operator-lib/pull/105/commits/3f60543cf9ce9403d249d252ce3cf1bd9a42d34d

everettraven:
As of now 1 test is expected to fail due to an issue with the controller-runtime `fakeClient` not respecting dry run options when running delete operations. I opened an upstream issue: https://github.com/kubernetes-sigs/controller-runtime/issues/1871

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
 
-	if len(config.Namespaces) == 0 {
-		return fmt.Errorf("namespaces are required")
+// NewPruner returns a pruner that uses the given strategy to prune objects.
+func NewPruner(prunerClient client.Client, opts ...PrunerOption) Pruner {
+	pruner := Pruner{
+		Registry: defaultRegistry,
+		Client:   prunerClient,
+		Logger:   Logger(context.Background(), Pruner{}),
+		GVK:      corev1.SchemeGroupVersion.WithKind("Pod"),
```
fgiloux, line 136: I am a bit uneasy with the defaulting. If GVK, namespaces and labels are not set by the caller, who is relying on the default values, all the Pods of all the namespaces are candidates for deletion. DefaultPodIsPrunable will thankfully limit that to Pods, which have succeed.

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
+			p.Logger.Info(fmt.Errorf("object is unprunable: %w", err).Error())
+			continue
+		} else if err != nil {
+			return nil, err
 		}
+
+		objs = append(objs, obj)
+	}
+
+	objsToPrune, err := p.Strategy(ctx, objs)
```
fgiloux, line 177: I haven't seen Strategy being defaulted like the other fields. It may not been guaranteed at compile time that it is not null as it is not an explicit parameter of NewPruner either

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
 
-		err = config.removeResources(ctx, resourcesToRemove)
-		if err != nil {
-			return err
-		}
-	}
+// IsPrunableFunc is a function that checks the data of an object to see whether or not it is safe to prune it.
+// It should return `nil` if it is safe to prune, `Unprunable` if it is unsafe, or another error.
+// It should safely assert the object is the expected type, otherwise it might panic.
+type IsPrunableFunc func(obj client.Object) error
```
fgiloux, line 83: What do you think of passing the logger  to the function?

fgiloux (commented):
lots of good things

fgiloux:
> I think we need to discuss the `Pruner` struct. Using functional arguments AND exposing fields kind of strikes me as an anti-pattern.

One option would be to completely unexpose the `Pruner` struct and to expose an interface with the Prune method instead, similar to what is done [here](https://github.com/kubernetes/client-go/blob/28ccde769fc5519dd84e5512ebf303ac86ef9d7c/informers/factory.go#L55-L67) and [here](https://github.com/kubernetes/client-go/blob/28ccde769fc5519dd84e5512ebf303ac86ef9d7c/informers/factory.go#L187)

prune/prune.go
everettraven: Thanks for bringing this up, @ryantking and I spoke offline a little bit about this yesterday and decided that it would be best to require the GVK be passed in as a parameter to the `NewPruner` function. I am currently working on making this change now.

prune/prune.go
everettraven: Thanks for bringing this up as well. Similar to my response above, Ryan and I spoke about this offline yesterday and we are adding the Strategy to the explicit parameters of NewPruner

prune/prune.go
everettraven: I'm not sure that is necessary since we are planning on using the logger in the `Pruner`'s `Prune` function to log when an object is not prunable. 

Is there a use case that you have in mind that might require the logger being passed into an `IsPrunableFunc`?

fgiloux: My thinking is that IsPrunableFunc will usually be implemented by the library users. It is difficult to anticipate what logic will be built in there and whether they want to log states that are noteworthy.

everettraven: That is a fair point, I don't think it hurts to pass it so might as well. Will work on making this change.

prune/prune.go
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/67688741005b777c097f0c3952c3d6c010d15db9

prune/prune.go
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/67688741005b777c097f0c3952c3d6c010d15db9

prune/prune.go
everettraven: done in https://github.com/operator-framework/operator-lib/pull/105/commits/67688741005b777c097f0c3952c3d6c010d15db9

prune/prune_test.go
```diff
@@ -0,0 +1,679 @@
...
+	)
+	BeforeEach(func() {
+		testScheme, err := createSchemes()
+		Expect(err).Should(BeNil())
+
+		fakeClient = crFake.NewClientBuilder().WithScheme(testScheme).Build()
+		fakeObj = &corev1.Pod{}
+
+		// Create our function to configure our pruner
+		prunerConfig = func(p *Pruner) {
```
fgiloux, line 57: Why using a new PrunerOption function rather than the existing withNamespace and withLabels?
My concern is that you can do that as the test is part of the prune package but library users would not be able to do the same as Pruner fields are not exposed.

fgiloux:
@everettraven it looks good to me. One question: would it make sense to provide a couple of most common strategies? What I have in mind:
- keep at least the latest X resources
- keep at least the resources newer than Y

prune/prune_test.go
everettraven: That's a good point. I made that before we had the existing `WithNamespace` and `WithLabels` options and it didn't cross my mind to change it. Thanks for bringing it up @fgiloux !

everettraven:
> @everettraven it looks good to me. One question: would it make sense to provide a couple of most common strategies? What I have in mind:
> 
>     * keep at least the latest X resources
> 
>     * keep at least the resources newer than Y

I think if we wanted to do something like this it would require more discussion as to what some sensible defaults are. 

My thoughts on it:
I'm a little hesitant to create some provided strategies that follow the `StrategyFunc` type and aren't very configurable and more so leave it to the library users to define their strategy.

As an example - if we took the idea of keeping at least the latest X resources and created a common strategy for it, we would have to explicitly define what X is within the logic of that common strategy. If a user wanted to use that common strategy but wanted to change X they wouldn't be able to.

```go
func KeepLatestXStrategy(ctx context.Context, objs []client.Object) ([]client.Object, error) {
	// in this case we would want to keep the 10 latest resources
	// there isn't currently a way to pass something like the number of resources you would like to keep in.
	x := 10

	// Just for an example return the first 10
	if len(objs) < x {
		return objs, nil
	}

	return objs[:x], nil
}
```

One thing I think we could try is to create some kind of helper function that returns a common strategy with a configurable value kind of like this:

```go
func GetKeepLatestXStrategy(x int) StrategyFunc {
	return func(ctx context.Context, objs []client.Object) ([]client.Object, error) {
		// Just for an example return the first 10
		if len(objs) < x {
			return objs, nil
		}
		return objs[:x], nil
	}
}
```

What do you think @fgiloux @ryantking ?

fgiloux:
@everettraven I like:

~~~
func GetKeepLatestXStrategy(x int) StrategyFunc {
	return func(ctx context.Context, objs []client.Object) ([]client.Object, error) {
...
~~~

it allows the users for simple use cases to have pruning implemented with ~4 lines of code.

prune/prune_test.go
```diff
@@ -210,7 +210,7 @@ var _ = Describe("Prune", func() {
 					Expect(len(pods.Items)).Should(Equal(1))
 				})
 
-				It("Should Prune Jobs with Default IsPrunableFunc", func() {
+				It("Should Prune Jobs with Default IsPrunableFunc", func() { //nolint:dupl
```
camilamacedo86, line 172: why the code is duplicated?
Could we not solve it instead of ignoring the lint check?
Also, why do we have more than one test with the same description name?

everettraven: There shouldn't be more than one test with the same description name. One should be "Should Prune Pods with Default IsPrunableFunc" and one should be "Should Prune Jobs with DefaultIsPrunableFunc".

I will create a function that both the tests can use since they pretty much do the exact same thing but use slightly different values.

everettraven: Should be resolved in https://github.com/operator-framework/operator-lib/pull/105/commits/264c78d7aa3a87b4191cff6e09f0a7d1b94cb08e

everettraven:
@fgiloux @ryantking Some common strategy functions should be added with commit https://github.com/operator-framework/operator-lib/pull/105/commits/3bf051ac2a5a2b6a9141c8d1c43c661e0d1e2257

Let me know what you think!

prune/strategies.go
```diff
@@ -0,0 +1,62 @@
...
+	return func(ctx context.Context, objs []client.Object) ([]client.Object, error) {
+
+		if len(objs) <= count {
+			return nil, nil
+		}
+
+		// sort objects by creation date
+		sortedObjs := objs
+
+		sort.SliceStable(sortedObjs, func(i, j int) bool {
```
fgiloux, line 38: Do we need stable sorting?

fgiloux (commented):
This looks good to me. I am just wondering whether we could save a few CPU cycles by not doing stable sorting.

prune/strategies.go
everettraven: We could probably get away without doing sorting, but sorting based on the date was my first thought that I felt would work well. Thinking on it some more, we may be able to do something like:
```go
var objsRemove []client.Object

for len(objs) - len(objsRemove) > count {
    // Loop through all the objs to find the oldest one.
}
```
However I feel like ^ would be less performant than sorting.

If we are more so concerned about using the native sort library that Go has because it uses reflect and is slower I could work on writing a different sort function that can be used to be a little bit more performant.

Any thoughts?

fgiloux: Sorry I was not very clear. I guess that there are good reasons for using reflect in the core go library. My question was not on "sorting" but "stable", i.e. whether we could use [sort.Slice](https://pkg.go.dev/sort#Slice) rather than [sort.SliceStable](https://pkg.go.dev/sort#SliceStable)

everettraven: Ah okay. I didn't think that much about the differences between the two. After taking a look at the differences, i feel like it would be fine to use the `sort.Slice` and save a few CPU cycles like you said. Thanks for bringing that up!

everettraven: I made a change to use `sort.Slice` in commit https://github.com/operator-framework/operator-lib/pull/105/commits/230e74e5b2422fe95863c28117c52bbeef99915c

fgiloux:
It looks good to me. Let see what @ryantking wants to add

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
-	_, err = labels.Parse(config.LabelSelector)
-	if err != nil {
-		return err
+	return pruner, nil
+}
+
+// Prune runs the pruner.
+func (p Pruner) Prune(ctx context.Context) ([]client.Object, error) {
+	var objs []client.Object
+	p.logger.Info("Starting the pruning process...")
```
ryantking, line 149: Nit:
```suggestion
	p.logger.Infof("pruning objects of GVK: %s", p.gvk)
```

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
+
+	return objsToPrune, nil
 }
 
 // Logger returns a logger from the context using logr method or Config.Log if none is found
 // controller-runtime automatically provides a logger in context.Context during Reconcile calls.
 // Note that there is no compile time check whether a logger can be retrieved by either way.
 // keysAndValues allow to add fields to the logs, cf logr documentation.
-func Logger(ctx context.Context, cfg Config, keysAndValues ...interface{}) logr.Logger {
+func Logger(ctx context.Context, pruner Pruner, keysAndValues ...interface{}) logr.Logger {
```
ryantking, line 196: I don't think this should be part of the exposed API. There should at most be a `pruner.Logger` function that returns the configured logger.

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
+	return fmt.Sprintf("unable to prune %s: %s", client.ObjectKeyFromObject(*e.Obj), e.Reason)
+}
+
+// StrategyFunc takes a list of resources and returns the subset to prune.
+type StrategyFunc func(ctx context.Context, objs []client.Object) ([]client.Object, error)
+
+// IsPrunableFunc is a function that checks the data of an object to see whether or not it is safe to prune it.
+// It should return `nil` if it is safe to prune, `Unprunable` if it is unsafe, or another error.
+// It should safely assert the object is the expected type, otherwise it might panic.
+type IsPrunableFunc func(obj client.Object, logger logr.Logger) error
```
ryantking, line 83: We should not include the logger in the signature here. What we have now couples the API to the `logr` package, which is something we very much don't want since there are many go logging options, plus it is likely that one day [go will have its own builtin structured logging interface](https://github.com/golang/go/issues/13182). Users who want a logger in their `IsPruneableFunc` should accomplish this with a closure:

```go
func MyLoggingIsPruneable(logger logr.Logger) IsPruneableFunc {
    return func(obj client.Object) error {
        logr.Info("I love logging!")
    }
}
```

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
+		p.namespace = namespace
 	}
+}
 
-	return false
+// WithLogger can be used to set the Logger field when configuring a Pruner
+func WithLogger(logger logr.Logger) PrunerOption {
+	return func(p *Pruner) {
+		p.logger = logger
+	}
```
ryantking, line 99: After talking with @everettraven we came to the conclusion that we can and should remove the logger from the pruner configuration. None of the internal pruning logic actually uses the logger so we can avoid coupling the library to any logger implementation. Users who want logging can implement their own is-pruneable and strategy functions that do whatever logging they want. They can even wrap the existing is-pruneable functions to add logging:

``` go
func myPodIsPruneable(obj client.Client) error {
    if err := prune.DefaultPodIsPruneable(obj); if prune.IsUnprunable(err) {
        // log some fun stuff
    } else if err != nil {
        // log some less-fun stuff
        return err
    }
    
    return nil
}
```

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
-		log = cfg.Log
+	if pruner.logger != (logr.Logger{}) {
+		log = pruner.logger
 	} else {
 		log = ctrllog.FromContext(ctx)
 	}
 	return log.WithValues(keysAndValues...)
 }
+
+func isUnprunable(target error) bool {
```
ryantking, line 206: Let's export this so users can check for `Unprunable` in their custom functions.

.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Run golangci-lint
-      uses: golangci/golangci-lint-action@v2
+      uses: golangci/golangci-lint-action@v3
```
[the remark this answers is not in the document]
everettraven: My understanding when doing some research was that the `golangci-lint-action@v2` will try to install Go rather than using the already installed one in the previous step that uses `actions/setup-go@v2`. In this case the `golangci-lint-action@v2` seemed to be trying to pull Go 1.18 for linting. Using `golangci-lint-action@v3` allows us to use the Go installed in the previous setup making sure that we are using the same go version.

prune/prune.go
everettraven: All logging implementation is removed with commit https://github.com/operator-framework/operator-lib/pull/105/commits/be7adfb415a270d4b5483d58e3d3a0abfc157934

prune/prune.go
everettraven: Should be exported as of commit https://github.com/operator-framework/operator-lib/pull/105/commits/7fb8237479526529abdc4dfabe4a4cc30ba4028b

prune/prune.go
```diff
@@ -16,184 +16,209 @@ package prune
...
 
-	if config.CustomStrategy == nil && config.Strategy.Mode == CustomStrategy {
-		return fmt.Errorf("custom strategies require a strategy function to be specified")
+// Namespace returns the namespace that the Pruner is using to find resources to prune
+func (p Pruner) Namespace() string {
+	return p.namespace
+}
+
+// NewPruner returns a pruner that uses the given strategy to prune objects that have the given GVK
+func NewPruner(prunerClient client.Client, gvk schema.GroupVersionKind, strategy StrategyFunc, opts ...PrunerOption) (Pruner, error) {
```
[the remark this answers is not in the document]
everettraven: Should be updated as of commit https://github.com/operator-framework/operator-lib/pull/105/commits/f881c0d3295f4e50df7969dec1e6ef06631d8671

fgiloux:
> @fgiloux Can you approve the EP? [operator-framework/enhancements#109](https://github.com/operator-framework/enhancements/pull/109) We need that merged to move forward with this. I have updated it to match the implementation here.

done (as much as I can)

everettraven:
@fgiloux @jmrodri @ryantking if you could re-review the most up to date changes we can try to have this ready to go right after the EP merges.

ryantking (approved):
/lgtm
/hold
Want signoff from @fgiloux then I'll unhold (or you can if you have the permissions)

jmrodri (approved):

everettraven:
@fgiloux When you have some time would you mind reviewing the latest changes?

fgiloux:
@everettraven sorry I am very busy these days. I am fine with the approach and what I have seen but I would like to trial the library to make sure I am not missing something in terms of UX. This requires unfortunately a bit of time.

PR #106, opened 2022-04-15
Closes #102: Make max backoff interval configurable

### LINKED ISSUE #102: Configurable max backoff interval in leader-for-life
## Feature Request

**Is your feature request related to a problem? Please describe.**
Yes. I found that the hard-coded max backoff interval of 16s was too long for production environments. It introduced unnecessary leaderless time during deployment.

**Describe the solution you'd like**
Make [maxBackoffInterval](https://github.com/operator-framework/operator-lib/blob/main/leader/leader.go#L47) configurable, which can be passed in via Become's [Option](https://github.com/operator-framework/operator-lib/blob/main/leader/leader.go#L50).

caueasantos:

Closes #102 
This pull requests aims to give users the ability of specifying custom duration intervals between attempts to become the leader. I tried to keep this change extremely simple exporting a new field to the become's Option.

The 16s hard-coded interval between attempts might cause unnecessary leaderless time during deployment. For example, we have a homemade crons solution responsible for handling jobs scheduling and we noticed a gap of jobs that have not been triggered here and there throughout the day as a result of a few seconds without a leader. Reducing the interval to become a leader should consistently prevent that from happening.


caueasantos:
@jmrodri Have I done something wrong? The CI errors seems not related to the changes and I'm not too sure how to move forward. What do you suggest?

caueasantos:
@jmrodri Can I get a review of this please? 🙏🏼

jmrodri (approved):
/lgtm

varshaprasad96 (approved):
/lgtm
@caueasantos Thanks for the PR! The change looks good. Just curious on the reason why you are using leader election from this repository? This follows the leader for life approach, which we may deprecate (not sure of this) in near future, since SDK has moved to using controller-runtime's leader for lease.

caueasantos:
> /lgtm @caueasantos Thanks for the PR! The change looks good. Just curious on the reason why you are using leader election from this repository? This follows the leader for life approach, which we may deprecate (not sure when) in near future, since SDK has moved to using controller-runtime's leader for lease.



> /lgtm @caueasantos Thanks for the PR! The change looks good. Just curious on the reason why you are using leader election from this repository? This follows the leader for life approach, which we may deprecate (not sure when) in near future, since SDK has moved to using controller-runtime's leader for lease.

No specific reason 😄 I'm going to take look and see if we can migrate it. Thank you

PR #108, opened 2022-05-24
(chore): bump k8s to 1.24

everettraven:
- bump Kubernetes dependencies to k8s 1.24

- remove workaround for DryRunClient in the prune library tests due to fix in controller-runtime

- replace deprecated clock library with the new one in the conditions library tests

Supporting k8s 1.24


camilamacedo86 (commented):
Great 🥇 

PS.: we need to upgrade kubebuilder otherwise we are unable to use this new version with SDK and bump the others libs

/lgtm

conditions/conditions_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
camilamacedo86, line 27: @everettraven Have we ci tests using go here? 
Could we also bump 1.18 on the tests only? So that we can ensure that we can run and it works with 1.18?

camilamacedo86: /lgtm cancel

^

camilamacedo86: It can be in a follow up as well 👍 
So, 
/approved

everettraven: I agree, I think it would be best to do this in a follow up when we upgrade to Go 1.18

rashmigottipati (approved):
/lgtm

PR #112, opened 2023-03-06
Update k8s dependencies to v1.26.1

rashmigottipati:
Made changes to bump k8s dependencies to support k8s 1.26.1. 
Changes:
- bump dependencies in go.mod to support k8s 1.26.1
- changes in go.sum from running `go mod tidy`

Support for k8s 1.26.1


go.mod
```diff
@@ -6,14 +6,14 @@ require (
...
+	github.com/operator-framework/api v0.17.4-0.20230223191600-0131a6301e42
 	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/client_model v0.3.0
-	k8s.io/api v0.26.0
-	k8s.io/apimachinery v0.26.0
-	k8s.io/client-go v0.26.0
-	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448
-	sigs.k8s.io/controller-runtime v0.14.1
+	k8s.io/api v0.26.1
+	k8s.io/apimachinery v0.26.1
```
grokspawn, line 13: Had a Q from Lucky's PR about why not 0.26.2 for apimachinery.  Do we have a position on this we can express?

go.mod
```diff
@@ -6,14 +6,14 @@ require (
...
+	k8s.io/client-go v0.26.1
+	k8s.io/utils v0.0.0-20230220204549-a5ecb0141aa5
+	sigs.k8s.io/controller-runtime v0.14.4
```
varshaprasad96, line 16: could we make this 0.14.5, since we are on it. The patch release reverts a few changes which may not necessarily affect here directly, but its better to use the latest patch with all the changes.

go.mod
varshaprasad96: Latest patch releases are always better, since K8s 1.26.2 does fix a few [regressions](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.26.md#changelog-since-v1261). Though it mayn't affect anything here directly, its better to use the latest ones to be on the safe side.

rashmigottipati: Makes sense. That sounds good to me! I was a bit hesitant as I thought we should be keeping it consistent across all the repos, but if that's not the case, I agree with your point. 

Bumped k8s to latest patch release `v0.26.2`

go.mod
rashmigottipati: Yup! Bumped controller-runtime to `v0.14.5` as well

rashmigottipati:
@varshaprasad96 - this is ready for review. Could you please take a look? 

/cc @everettraven @grokspawn

varshaprasad96 (approved):
/lgtm
/approve

PR #121, opened 2023-10-12
Bump go version to v1.20

rashmigottipati:


rashmigottipati:
cc @oceanc80 @everettraven

everettraven (approved):

everettraven (approved):
/lgtm

varshaprasad96:
@rashmigottipati looks like golangci-lint may be failing because of cache issues. I haven't dug more into this - but probably cache is enabled by default, which is why running `setup-go` multiple times is causing it to fail for every action. If a re-trigger does not fix it (just to ensure its not a flake), we can disable cache as a work around (https://github.com/golangci/golangci-lint-action/issues/807). We should probably be opening an issue to investigate further on why this is happening, as disabling cache could make our CI runs slow.

joelanford:
This is a golangci-lint thing. It needs its version bumped to handle the new go version. You need at least version 1.51.0.

everettraven:
>/test lint

@oceanc80 I think that is a github action test, not one managed by prow. I'll close and re-open to try and kick tests and have them pick up the recent changes

PR #126, opened 2024-01-10
Enable ginkgolinter and fix findings

nunnatsa:
## Description of the change
ginkgolinter finds bugs and enforces standards of using the ginkgo and gomega packages. See more details here:
https://github.com/nunnatsa/ginkgolinter

This PR enables the ginkgolinter in the .golangci.yml, and fixes all the new finding from running golangci-lint. It also sorts the enabled linter names for better readability.

Note: all the finding were auto fixed by running the ginkgolinter cli with the `-fix` flag.

While going over the changes, found several more issues:

First, the `assertMetrics` function in handler/instrumented_enqueue_object_test.go, wasn't really doing anything. The switch-case compared a pointer value to another pointer, so the no case was actually ever selected, because the addresses were not the same.

This PR also fixes that issue.

Then, this PR also improve the error handling in tests in two ways:
Replace this pattern:
```golang
err := someFuncRetOnlyErr()
Expect(err).ToNot(HaveOccurred()
```
With
```golang
Expect(someFuncRetOnlyErr()).To(Succeed())
```

Also, use the `MatchError` gomega matcher when checking errors.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
   - asciicheck
   - bodyclose
   - errorlint
+  - ginkgolinter
   - gofmt
   - goimports
   - gosec
--- a/conditions/conditions_test.go
+++ b/conditions/conditions_test.go
@@ -64,25 +64,21 @@ var _ = Describe("Condition", func() {
 			}
 
 			// Create Operator Condition
-			err := os.Setenv(operatorCondEnvVar, "operator-condition-test")
-			Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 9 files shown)


ncdc:
@nunnatsa thanks for the PR! Let's please coordinate - I have #125 in this space too

nunnatsa:
> @nunnatsa thanks for the PR! Let's please coordinate - I have #125 in this space too

Hi. I'll need to rebase anyway, as #125 was merged.

But does the pr makes sense?

ncdc:
Thanks, I'm afk for a bit, but will review when I'm back today

ncdc (approved):

PR #157, opened 2024-03-28
Release leader for life lock in case pod is preempted.

jagpreetstamber:


jhmadhav (approved):

leader/leader.go
```diff
@@ -179,8 +179,8 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 					log.Info("Leader pod has been deleted, waiting for garbage collection to remove the lock.")
 				case err != nil:
 					return err
-				case isPodEvicted(*leaderPod) && leaderPod.GetDeletionTimestamp() == nil:
-					log.Info("Operator pod with leader lock has been evicted.", "leader", leaderPod.Name)
+				case (isPodEvicted(*leaderPod) || isPodPreempted(*leaderPod)) && leaderPod.GetDeletionTimestamp() == nil:
+					log.Info("Operator pod with leader lock has been evicted or preempted.", "leader", leaderPod.Name)
```
everettraven, line 183: Nit: Instead of running this case when either the pod is evicted _or_ preempted, I think we should make a separate case statement for the scenario where the pod is preempted. Having a separate case allows for better log messages that state explicitly the scenario that occurred (no questioning whether the pod was `evicted` or `preempted` based on reading the logs) and allows us to change the behavior for each case independently if we see fit in the future.

everettraven (approved):
Thanks @jagpreetstamber for the PR! I had one nit that I left as a comment, but it's more of a suggestion - feel free to make the change or leave as is.

leader/leader.go
jagpreetstamber: Resolved in f4b850e

everettraven (approved):
/lgtm

Issue #117, opened 2023-07-10
Deprecate Leader for life based leader election


varshaprasad96:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Currently, the repository contains a leader-for-life based election model that ensures that a single leader is elected for life during a HA state.

To briefly describe what leader for life and leader for lease approaches are:
(1) Leader for life: A leader pod is selected for life until its garbage collected. This ensures there is only one leader at any instant of time.
(2) Leader for lease: Leader election happens periodically at a defined interval which can be tweaked. When the current leader is not able to renew the lease, a new leader is elected.

More details on both of these approaches are available [here](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#leader-election). 

Approach (2) is implemented upstream, in [client-go](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection) and is scaffolded by default for the past 30+ releases of SDK, since it is implemented as a part of setting up the manager in controller-runtime. It guarantees faster election of leaders, less downtime, recovery from disconnected/frozen node failures. However, it does not eliminate the split brain scenario - where more than a single leader is available at an instant of time.

Approach (1) on the other hand, was developed long before we had a leader-for-lease implemented upstream. Though it solves the split brain scenario, it does not guarantee recovery from node failure nor faster recovery. We also have issues with integrating it to controller-runtime (https://github.com/operator-framework/operator-lib/issues/48). Neither is it being maintained nor used as widely as leader for lease.

Here is a detailed [comment](https://github.com/operator-framework/operator-sdk/issues/784#issuecomment-442549839) explaining the preference of (2) over the other, wherein users would prefer a faster recovery even though there is split brain scenario intermittently, rather than an implementation that does not guarantee faster recovery. 

**Describe the solution you'd like**
Since leader for life approach is not being widely used, neither works with controller-runtime seamlessly, it is better to adopt a well tested upstream library than to depend on what is currently available as an option. 

The solution for this is:
1. To deprecate and remove leader for life in future releases of Operator SDK.
2. To bring this up upstream (in controller-runtime), for easier integration. This had already been brought up upstream (https://github.com/kubernetes-sigs/controller-runtime/issues/1963) but there was no response on the same.

PR #163, opened 2024-04-18
Bump k8s to 1.29

everettraven:
- Bumps k8s libraries to 0.29.3
- Bumps controller-runtime to 0.17.3
- Bumps o-f/api to 0.23.0
- Bumps Go to 1.22
- Updates `leader` package test logic to fix a test failure related to the GVK and type meta not being set after performing a GET with a controller-runtime fake client

- Stay up to date with kubernetes bumps happening in Operator Framework projects


leader/leader.go
```diff
@@ -291,6 +292,10 @@ func getNode(ctx context.Context, client crclient.Client, nodeName string, node
 		log.Error(err, "Failed to get Node", "Node.Name", nodeName)
 		return err
 	}
+	// .Get() clears the APIVersion and Kind,
+	// so we need to set them before returning the object.
+	node.TypeMeta.APIVersion = "v1"
```
stevekuznetsov, line 297: Having to mess with TypeMeta manually is a bit of a smell - why did we need this?

leader/leader.go
```diff
@@ -174,6 +174,7 @@ func Become(ctx context.Context, lockName string, opts ...Option) error {
 				leaderPod := &corev1.Pod{}
 				key = crclient.ObjectKey{Namespace: ns, Name: existingOwners[0].Name}
 				err = config.Client.Get(ctx, key, leaderPod)
+				leaderPod.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("Pod"))
```
stevekuznetsov, line 177: Ditto to setting GVK - what are we doing that requires it? Can we use a Scheme with the kinds registered?

Makefile
```diff
@@ -20,7 +20,7 @@ fmtcheck: ## Check go formatting
 	@gofmt -l $(SOURCES) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
 
 test: ## Run unit tests
-	@go test -race -covermode atomic -coverprofile cover.out ./...
+	@go test -timeout=60s -v -race -covermode atomic -coverprofile  cover.out ./...
```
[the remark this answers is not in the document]
everettraven: Fair point, this is left over from my local debugging. Meant to clean this up and missed it

leader/leader.go
everettraven: I made this change to satisfy the failing test that was expecting this info to be set on the object after the `config.Client.Get()` call but it wasn't. Didn't think to look into setting the Scheme - will give that a shot. 

I was trying to go for a minimal change to existing behavior and follow precedence of manually setting stuff after a Get call since I _believe_ this package has been for a while intended to be deprecated (see #117 - I could be misinterpreting this though) and the intent was to get the kubernetes packages bumped.

leader/leader.go
everettraven: Did this for precedence and a test that used the TypeMeta information: https://github.com/operator-framework/operator-lib/blob/3bfabe03dfae8754ef3a3dc02b09ec618e2ecb78/leader/leader.go#L277-L280

Makefile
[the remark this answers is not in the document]
everettraven: Reverted in https://github.com/operator-framework/operator-lib/pull/163/commits/c5522da9cd8a162923102c14bdeaa5d2b58f5499

leader/leader.go
everettraven: No scheme changes were necessary - I totally missed it, but looked like the tests were using a fake client. Setting the appropriate fields on the objects in the fake client resolved the issue.

jberkhahn:
Looks like it's failing the linter? I definitely see a bunch of extraneous spaces that could be removed

Neo2308:
@everettraven missed your PR and had created a new one for these bumps: https://github.com/operator-framework/operator-lib/pull/167
Included your fixes over there, but didn't do the golang bump to 1.22 (shouldn't that be done when we bump to k8s 1.30?)

PR #183, opened 2024-06-20
Fix codecov-action params

m1kola:

* `functionalities` param is no longer exist. It was used to enable file fixes to ignore common lines from coverage. This feature is now seems to be on by default.

* Adding `disable_search` because we do not need for the codecov action to search for coverage files: we explicitly provide files.

Fixing codecov config


everettraven (approved):
/approve
/lgtm

